PR #86, opened 2022-03-25
Edit docs for fig extension and typo

mathren:
Indeed LaTeX does not need the extension if using the graphics package (https://docs.aspose.com/tex/net/latex-graphics/), but since showyourwork does, I think it is worth mentioning in the documentation.


rodluger:
Thanks!

PR #88, opened 2022-04-18
Hack to ignore orcid-id file in other_filenames

adrn:
In the [astropy paper](https://github.com/astropy/astropy-v5.0-paper) build, for some reason the orcid-ID.png file was getting picked up by this rule and added to `other_filenames` (as many times as I have authors in my manuscript file). This does two things: switches to using a `set()`, because I don't think there's a reason to have duplicates in `other_filenames`, and adds a hack to check if the file considered is the orcid-ID file (and if so ignores it). I know this is a brittle change / hack, but it works for me locally. And I know this repo is deprecated, but if we can get some fix in for this it would be easier than switching over to the new system (for this paper)...

### DIFF
--- a/workflow/rules/script_info.smk
+++ b/workflow/rules/script_info.smk
@@ -76,8 +76,8 @@ checkpoint script_info:
                             "{}.{}".format(relpaths.figures / label, ext), []
                         )
                         datasets = [
-                            dataset for dataset in alldeps 
-                            if type(dataset) is str and 
+                            dataset for dataset in alldeps
+                            if type(dataset) is str and
                             dataset.endswith(".zenodo")
                         ]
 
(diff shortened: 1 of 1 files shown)


rodluger:
Thanks for this! The orcid-ID file should be caught in [this](https://github.com/rodluger/showyourwork/pull/88/files#diff-616ec5f4e2fea3eef8f4674cb18db2614d8c84f8a1b82fe8890d253bdb1adc13R124) block, since it's defined in "special_figures". I don't quite understand why it's being caught in [this](https://github.com/rodluger/showyourwork/pull/88/files#diff-616ec5f4e2fea3eef8f4674cb18db2614d8c84f8a1b82fe8890d253bdb1adc13R120) block, since it shouldn't be nested within the `figures` directory (it should get placed directly alongside the manuscript). But if these changes work for you, I'm happy to approve them -- doesn't look like they would cause adverse effects. I'll run the tests and release a new version shortly.

PR #91, opened 2022-05-10
Adding support for module level execution

dfm:
This PR adds support for executing the showyourwork CLI using the following syntax

```bash
python -m showyourwork ...
```

taking the same arguments as the script-like

```bash
showyourwork ...
```

version. This can be useful for environments with path collisions, and I just pretty much always like to run Python scripts this way to make sure that I know which Python installation they are using.


rodluger:
Thanks!

PR #94, opened 2022-05-18
A few very minor updates to the CLI text 

adrn:
As you were live demoing, I had some ideas for clarifying and making the CLI a little nicer to users. Here specifically, I changed out the accusatory language like "you didn't enable X" and replaced with text explaining that in some cases "not enabling something" is just the default behavior.


rodluger:
Thanks!

PR #95, opened 2022-05-20
Added VSCode instructions to FAQ

jfcrenshaw:
I updated the VSCode instructions, including new instructions so that "Autobuild on save" works with `showyourwork`.


rodluger:
Awesome, thanks for contributing this!

Issue #104, opened 2022-06-03
renaming of manuscript: possibly wrong example in the docs?


mathren:
Hi Rodrigo! Thanks for the work you are putting in this great tool!

I've been trying to create a new repo, and create a renamed manuscript (I really don't want to have multiple `ms.tex` open in my text editors for different projects).  Following the [docs](https://show-your.work/en/latest/config/?highlight=rename#ms), I've added to `showyourwork.yml` the line ```ms:  src/tex/article.tex``` on a brand new template repository, and renamed `ms.tex` to `article.tex`, which appears correctly if i `ls` the folder `src/tex`. However, upon running `showyourwork` I hit the error:

```
FileNotFoundError in line 43 of /mnt/home/mrenzo/.showyourwork/env/lib/python3.9/site-packages/showyourwork/workflow/prep.smk:
[Errno 2] No such file or directory: '/mnt/home/mrenzo/Projects/CE_accretors/src/tex/ms.tex'
  File "/mnt/home/mrenzo/.showyourwork/env/lib/python3.9/site-packages/showyourwork/workflow/prep.smk", line 43, in <module>
  File "/mnt/home/mrenzo/.showyourwork/env/lib/python3.9/site-packages/showyourwork/config.py", line 274, in parse_config
  File "/mnt/home/mrenzo/.showyourwork/env/lib/python3.9/site-packages/showyourwork/config.py", line 89, in get_class_name
```

where the second line suggests that despite the edit in the `showyourwork.yml`, it is still looking for `ms.tex`.

I am able to build if I edit what is line 222 in `showyourwork/config.py` (in version `0.3.0.dev8`) to be:
``` config["ms_name"] = config.get("ms_name", "article")``` instead of ```config["ms_name"] = config.get("ms_name", "ms")```, that is, if I hardcode the new name of the manuscript in `config.py`. This is clearly not the desired behavior. After messing around with `config.py` for a bit, I realized that actually *not changing anything* in `config.py` but changing the line in `showyourwork.yml` to be:

```
   ms_name: article
```
(with still the constrain that the tex file has to be in `src/tex` and named `article.tex`) works.
So unless `config.py` is not how you want it, the minimal change to get the renaming to work would be to change the docs to this. Happy to do the change myself if you approve!

rodluger:
Thanks for catching this! The docs are wrong, and your implementation is correct. Please do submit a PR if you have a second!

PR #106, opened 2022-06-07
change docs on manuscript name

mathren:
fix for https://github.com/showyourwork/showyourwork/issues/104#issuecomment-1148946514

I effectively just changed `ms` to `ms_name`. I also changed the default value on what is coded in `config.py`, i.e., `src/text/ms.tex` is just `ms`.


rodluger:
Thanks!

Issue #110, opened 2022-06-15
Question about paths.py


jfcrenshaw:
The `paths.py` module is really useful! But I wanted to organize my `scripts` directory into subdirectories, which created a problem, since my scripts couldn't automatically find `paths.py` anymore as they weren't in the same directory.

I could have added some `os` stuff at the top of my scripts to add the root of `scripts` to the path, but that felt hacky and clunky. I ended up adding `showyourwork` to my `environment.yml`, and now I start all of my scripts with
```
from showyourwork.paths import user as Paths

# instantiate the paths
paths = Paths()
```

There's no problem here - everything works fine. I was just wondering if you had any thoughts about this. Should this solution be added to the docs in case other people want to organize their `scripts` directory into subdirectories? Or is there a more obvious solution that I missed?

rodluger:
Haha, nice hack!

I didn't do that by default because there are already (potentially) two different versions of `showyourwork` in play when you build an article: the one installed in your current (base) environment and the one that's installed within the isolated environment that builds the article. (The workflow you are running defines its own version of `showyourwork` in the config file, which may be different than the version that gets imported when you run `showyourwork` on the command line, hence the need for two separate versions).

Having the workflows themselves depend on `showyourwork` brings a _third_ version of the package into play. Users can definitely set it to be the same version as that in the `showyourwork.yml` file, but it will still be installed in a separate, isolated conda environment.

There's nothing wrong with that -- I just didn't want to make that the default behavior since it feels cumbersome and could lead to any number of issues due to version incompatibility. (Having two versions of `showyourwork` in play, as it currently stands, already keeps me up at night sometimes).

But, long story short, what you're doing is fine! It's just useful (I think) to understand why I didn't implement it that way in the first place. Feel free to issue a PR to add this to the docs (perhaps under FAQs)?

Thanks!

jfcrenshaw:
Makes sense! I will add some help on this to the FAQ

PR #111, opened 2022-06-15
Added docs about using paths within scripts subdirectories.

jfcrenshaw:
Spawned from issue #110 

Not sure where you wanted this placed -- feel free to move it around!


rodluger:
Awesome, thanks!

Issue #117, opened 2022-06-27
cache publishing issues


mathren:
Hi Rodrigo,

I have a showyourwork with [public static dataset](https://zenodo.org/record/6600641) and a [cache](https://sandbox.zenodo.org/record/1074085) that are both public (repo isn't yet). Both locally and on github actions the pdf is successfully compiled *but* showyourwork is not putting the dataset links in the figure, e.g.:

![image](https://user-images.githubusercontent.com/49067756/175993884-347a4cec-2616-4442-8785-c5003da18fe5.png)

Moreover, despite having both the ZENODO_TOKEN and SANDBOX_TOKEN defined (and passed as secrets to the repo), `showyourwork cache publish` fails with an error 403 (while `create` and `freeze` worked fine). When I ran `showyourwork cache freeze` the sandbox repository went from `draft` to `published` on sandbox.zenodo.

rodluger:
Currently, showyourwork only adds the Zenodo link if the Zenodo dataset is a *direct* dependency of the figure. In your workflow, there's an intermediate step (the unzipping rule) between the dataset and the figure script, so the icon isn't added automatically. I can change the behavior of showyourwork to do what you expect in this case, though -- shouldn't be too hard.

rodluger:
Hey @mathren I fixed this as well. Can you try running your workflow again without making the dataset dependency of the figure explicit? In other words, I believe you have a Zenodo dataset `A` that gets unzipped in a custom rule to a file `B`, and `B` is marked as a dependency of your figure. Previously, this would **not** give you a Zenodo icon next to the figure, since the dependency of the figure on `A` was implicit. The current version on the `main` branch fixes this behavior, and it should infer the upstream dependency and give you an icon. Let me know if it works!

mathren:
mmh, that does not seem to work for me. using `0.3.1.dev16+ga77bd0a` I get the same behavior as before.
Also, I've noticed that the hyperlinks in the  icons of the pdf compiled by arXiv (from the showyourwork-generated tarball) are broken (possibly same/similar bug as the naming of the cache on zenodo).

[...]

PR #118, opened 2022-07-01
fix link for ZENODO_TOKEN

mathren:
The older version still pointed to the sandbox, causing the issue
https://github.com/showyourwork/showyourwork/issues/117
creating the proper token on zenodo (as opposed to sandbox), then caching works.

(real change on line 357, rest are white space that my editor decided to mess with).


rodluger:
Awesome, thanks for catching this!

PR #129, opened 2022-07-23
Optimize the dag for cached rules

### LINKED ISSUE #124: Fix behavior for cached rules that depend on Zenodo datasets
This is an issue identified by @katiebreivik. Snakemake evaluates whether or not to restore a rule's output from the cache based on the hash of all the upstream rules. This check, however, happens _after_ the workflow graph (the DAG) has been built. In `showyourwork`, if a given rule depends on files downloaded from a (static) Zenodo deposit, the current behavior means we must always download those files, regardless of whether the rule is actually executed or simply restored from the cache. This is computationally wasteful -- in @katiebreivik's case, there are hundreds of (very large) files that are used _only_ in that rule, so downloading those files---only to then decide they're not needed, as we don't have to run the rule at all---is silly.

rodluger:
Fixes #124.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ htmlcov
 junit
 **/sitecustomize.py
 dist/*
+.vscode
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -23,6 +23,11 @@
   accessing the remote file cache.
 - Fixed issue with broken GitHub links in the PDF when using SSH to authenticate
   with GitHub.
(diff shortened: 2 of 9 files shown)


rodluger:
@katiebreivik Any more tests you'd like to run, or can we merge this PR?

katiebreivik (approved):
I think we've stress tested this enough over [here](https://github.com/katiebreivik/hush/actions).

The docs also look good, so I think this is ready to go!

katiebreivik:
merge away! @rodluger

PR #140, opened 2022-08-03
Remove imports of `fontspec` package to fix tt fonts

MilesCranmer:
This should in principle fix the use of custom typewriter fonts when compiling with `tectonic`. @rodluger


rodluger:
Merge workflow action failed because of a server error on GitHub Actions. Re-running it.

PR #142, opened 2022-08-08
Fixing overleaf branch name

dfm:
Overleaf uses `master` as its branch name, but I have `main` set as the default name in my git config. This meant that I couldn't set up a new project with Overleaf integration since it would fail with an error saying that there was no local branch called `master` when trying to push. This change fixes things for me locally.


rodluger:
Nice catch, thanks!

PR #143, opened 2022-08-08
Relaxing conda channel settings

dfm:
The `defaults` conda channel is pretty trash for M1 macs and the default syw environments don't play nice with my setup (i.e. conda says it can't solve the requested dependencies). Removing the channel list and specifying the channels only when necessary while falling back on the system settings for the others works well for me and I don't think we lose anything with this change!


rodluger:
Cool, thanks! This will help a lot of people I imagine.

Issue #145, opened 2022-08-09
Handling multiple labels exception


dfm:
The Annual Reviews `documentclass` (#144) automatically inserts its own label into figures (I assume for table of contents reasons?) so we're getting the following error from showyourwork:

```
A figure has multiple labels: `fig:quasar, chk@figure@1`
```

Originating here:

https://github.com/showyourwork/showyourwork/blob/0c065b2dc5f74b9612ab064d6674b91e206404ce/showyourwork/workflow/scripts/preprocess.py#L210-L215

It's not obvious to me why we need to require only a single label. Would it be possible to soften this requirement?

rodluger:
Interesting. The figure label currently needs to be unique because it is the [key in the dictionary](https://github.com/showyourwork/showyourwork/blob/0c065b2dc5f74b9612ab064d6674b91e206404ce/showyourwork/workflow/scripts/preprocess.py#L417) mapping figures to scripts, datasets, cache links, etc. In principle it doesn't matter what that label is called, so if there are two labels in one figure environment, we could just map all the figure attributes to the first label and ignore the second one. We already kind of do that [here](https://github.com/showyourwork/showyourwork/blob/0c065b2dc5f74b9612ab064d6674b91e206404ce/showyourwork/workflow/scripts/preprocess.py#L329) -- so, like you said, it's really just a matter of getting rid of that exception.

The second issue is that the way we display margin icons next to figures is by hacking the `showkeys` package, which is used to display the `\label` corresponding to various manuscript elements in the margin. People use this during writing to, e.g., easily identify which equation label corresponds to a given equation number by just glancing at the PDF. `showyourwork` [overrides this functionality](https://github.com/showyourwork/showyourwork/blob/0c065b2dc5f74b9612ab064d6674b91e206404ce/showyourwork/workflow/resources/styles/build.tex#L197) to display the associated icons instead of just the label name (and only for figures, currently).

Again, in principle at least, this shouldn't be too much of a problem. One of the labels for the figure in your case won't have any associated icons, so [these if statements](https://github.com/showyourwork/showyourwork/blob/0c065b2dc5f74b9612ab064d6674b91e206404ce/showyourwork/workflow/resources/styles/build.tex#L201) will evaluate to false and display nothing. But I don't know how `showkeys` itself behaves when elements have multiple labels, so we'd have to test that.

[...]

PR #148, opened 2022-08-09
Adding option to use a fork of showyourwork

dfm:
I think that this should be backwards compatible, but it permits the following syntax:

```yaml
version:
  fork: https://github.com/dfm/showyourwork.git
  spec: some-branch-tag-or-commit-hash
```

Omitting the `spec` argument will use the default branch, and omitting the `fork` argument will use the base `showyourwork/showyourwork` repo.

Ref: #145 for some context

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -739,7 +739,7 @@ crank up the verbosity even more by passing the ``--verbose`` argument to
 ``version``
 ^^^^^^^^^^^
 
-**Type:** ``str``
+**Type:** ``str`` or ``mapping``
 
 **Description:** The version of the |showyourwork| package required to build
 the article, populated automatically when ``showyouwork setup`` is run. Users
@@ -748,6 +748,7 @@ package. Options are:
 
 - any pip-installable version number (e.g., ``0.3.0``)
 - a 5-character (short) or 40-character (long) GitHub commit SHA (e.g, ``abcde``) corresponding to a specific commit to the `showyourwork/showyourwork <https://github.com/showyourwork/showyourwork>`_ repo
+- a mapping with the ``fork`` and/or ``spec`` keys described below
 
 **Required:** yes
 
@@ -756,3 +757,44 @@ package. Options are:
 .. code-block:: yaml
 
   version: 0.3.0
+
+.. _config.version_fork:
+
+``version.fork``
+^^^^^^^^^^^^^^^^
+
+**Type:** ``str``
+
+**Description:** The URL of a GitHub fork of the |showyourwork| repo.
+
+**Required:** no
+
+**Default:** ``https://github.com/showyourwork/showyourwork.git``
+
+**Example:**
+
+.. code-block:: yaml
+
+  verbose:
+    fork: https://github.com/SOMEUSER/showyourwork.git
+
+.. _config.version_spec:
+
+``version.spec``
+^^^^^^^^^^^^^^^^
+
+**Type:** ``str``
+
+**Description:** The spec for a specific branch, tag, or commit on the fork of
+the |showyourwork| repo specified by ``version.fork``.
+
+**Required:** no
+
+**Default:** ``null``
+
+**Example:**
+
+.. code-block:: yaml
+
+  verbose:
+    spec: main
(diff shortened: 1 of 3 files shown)


dfm:
I can confirm that this seems to work as expected for my project, although I don't know how to include an integration test.

docs/config.rst
```diff
@@ -756,3 +757,44 @@ package. Options are:
...
+
+**Required:** no
+
+**Default:** ``https://github.com/showyourwork/showyourwork.git``
+
+**Example:**
+
+.. code-block:: yaml
+
+  verbose:
```
rodluger: typo here, should be `version`

docs/config.rst
```diff
@@ -756,3 +757,44 @@ package. Options are:
...
+
+**Required:** no
+
+**Default:** ``null``
+
+**Example:**
+
+.. code-block:: yaml
+
+  verbose:
```
rodluger: typo here, should be `version`

rodluger:
Cool, thanks! Just wondering: why not keep it simple and let the user provide something like

```
version: https://github.com/user/repo@spec
```

instead of the `fork:` and `spec:` logic?

dfm:
I think the regex logic in there already is pretty gnarly so this seemed easier to maintain in the long term, but I'd be happy to use that mode if you prefer!

rodluger:
I agree your approach is more elegant. It also opens the door to allowing arbitrary branch names in the showyourwork/showyourwork repo (currently I only allow `main` and `dev` since other branches could in principle look like relative path names on disk). This would let us do something like

```
version:
    branch: foobar
```

So it might be useful to take a step back now and figure out a better syntax for all possible version specifications:

- PyPI version
    ```
    version:
        pip: 0.3.0
    ```
- GitHub SHA
    ```
    version:
        sha: abcde
    ```
- GitHub branch
    ```
    version:
        branch: main
    ```
- GitHub SHA on fork
    ```
    version:
        fork: dfm/showyourwork # or github.com/dfm/showyourwork
        sha: abcde # or branch: main
    ```
- Local
    ```
    version:
        path: /path/to/local/folder
    ```

And, of course, all backwards-compatible version specs. So we'd allow either a string (current behavior), `pip`, or one of (`path`, `sha`, or `branch`). For `sha` and `branch`, we further allow an optional `fork` key that defaults to `showyourwork/showyourwork`.

What do you think?

dfm:
I think this is great - Happy to implement!

rodluger:
Yes please! As for testing, the current approach is fine. We can integrate it into the full testing workflow tomorrow when we chat.

rodluger:
(BTW, is the url structure `<baseurl>/user/repo.git@spec#egg=egg` universal? Is that the case for bitbucket, gitlab, etc.?)

dfm:
All of the above work now plus also `tag` which should have the same syntax. A different option would be to just use `ref` instead of `branch`, `tag`, and `sha` since they all have the same behavior and I think `ref` is the usual name...

dfm:
I think that this syntax would work for other `git`-based platforms, yes!

dfm:
I did switch to `ref` to cover `branch`, `tag` and `sha`. If you'd like to include those as aliases, we can add them back in.

rodluger:
Yeah, `ref` is good -- thanks!

rodluger:
BTW, I don't have Actions set up to run tests on PRs because the testing workflow is expensive and spins up other actions workflows in all the test repositories (`showyourwork/test-***`). Would be nice to allow the user to request a workflow run by, e.g., posting a PR comment and having a bot spin up the test after admin approval -- or something like that. What do you think @dfm?

dfm:
That would be very cool!! 🤖

rodluger:
@dfm can you rebase and commit to see if it triggers a workflow run?

rodluger:
Nice. The new strategy for running PR workflows is working! (In brief: I use a `pull_request_target` trigger, since the workflow needs to be able to write to the test repos in the `showyourwork` org, but I execute the workflow within an isolated [environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) that requires admin approval to run).

dfm:
Sweet!!

rodluger (approved):

rodluger:
Thanks for implementing this!

PR #151, opened 2022-08-10
Refactoring structure of tests

dfm:
Based on conversations with @rodluger, this PR refactors the testing infrastructure to make the developer experience a little nicer and to make future additions a little easier. The changes:

1. The tests are now separated into "unit tests" and "integration tests", with most of the old tests being integration tests since they test the full workflow.
2. A custom `pytest.mark.remote` is added to label integration tests as requiring the secrets that are only available in the main showyourwork repo. By default these tests are skipped, and they must be directly requested with the `--remote` command line argument to `pytest`.
3. The GitHub Actions infrastructure should now run the unit tests and non-remote tests on PRs by default, with the remote tests still requiring approval.

TODO:

- [x] Update `unit_tests.rst` docs


dfm:
Another note for @rodluger. Instead of using the deployment environment, I'm looking at using the `safe to test` label as suggested here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

dfm:
I think this one is ready to go. The only remaining question is whether or not the new `pull_request_target` label-based dispatch is sensible. I'm inclined to merge then test it out, but let me know what you think @rodluger

PR #158, opened 2022-08-11
Adding and running pre-commit for consistent style

dfm:
This looks like a huge PR, but I've just added a pre-commit config which enforces a consistent style across the project. We were pretty much already following this, so there aren't any functional changes.

@rodluger are you happy with including something like this? https://pre-commit.ci will automatically check and fix style errors on PRs.


rodluger (approved):
Looks good!

PR #161, opened 2022-08-13
Starting to work on bi-directional Overleaf support

### LINKED ISSUE #131: Bidirectional overleaf support
It would be really nice if you could edit TeX on both overleaf and a showyourwork repo, and have them sync with eachother. (I think this isn't needed for figures - all the figures+scripts can have their truth copy in the showyourwork repo.) One would need to manually fix merge conflicts, but I think this is a small price to pay for having a completely synchronized overleaf repo. Many collaborators still use overleaf so it would be nice to have a tighter integration and remove this final roadblock.

One option to do this is to use something like Google's "Copybara" which synchronizes open-source projects with Google's internal closed-source monorepo:

https://github.com/google/copybara

This might be overkill though; perhaps just commiting changes and having the user manually fix merge conflicts, rather than force pushing, is good enough?

dfm:
This is a WIP implementation of a bi-directional Overleaf implementation. So far I just have logic to handle bringing in changes from Overleaf when there have been local changes too. It will try to rebase automatically, but if it can't, the user is required to complete the rebase manually.

Notes to self:

- This doesn't currently handle things like figures that shouldn't be tracked in SYW. We can handle those using `--exclude` patterns when applying the initial diff.
- How do we handle pushing changes back to Overleaf? I think it's nearly identical, but we'll need to think about handling merge conflicts. We could save a temporary directory with the checked out Overleaf repo.
- Still need to think about how to handle migration from existing Overleaf integrations, and start up for new projects.
- What about the front end? I like the idea of adding a `showyourwork sync` command to execute this update, and perhaps an `showyourwork build --sync` command that would work like the current `showyourwork build`, syncing every time.

closes #131 (eventually)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ setup(
         "cookiecutter",
         "packaging",
         "lastversion",
+        "plumbum",
     ],
     extras_require={"tests": ["pytest", "pytest-asyncio-cooperative"]},
     entry_points={
--- a/showyourwork/cli/commands/sync.py
+++ b/showyourwork/cli/commands/sync.py
@@ -1,12 +1,9 @@
-
 from ... import logging, paths
 from ...config import edit_yaml
-from ...git import get_repo_branch
 from ...overleaf2 import Overleaf
 
 
-def overleaf_sync(init=False, force=False):
-    branch = get_repo_branch()
+def overleaf_sync(continue_=False, init=False, force=False):
     with edit_yaml("overleaf.yml") as config:
         version = config["overleaf"].get("version", 1)
         assert version == 2
@@ -16,3 +13,17 @@ def overleaf_sync(init=False, force=False):
         )
         local_sha = config["overleaf"].get("local_sha", None)
         remote_sha = config["overleaf"].get("remote_sha", None)
+
+    overleaf = Overleaf.from_url(
+        url, local_sha=local_sha, remote_sha=remote_sha
+    )
+
+    if init:
+        overleaf.setup_remote(force=force)
+        return
+
+    if continue_:
+        overleaf = overleaf.continue_sync()
+
+    else:
+        overleaf = overleaf.sync(force=force)
--- a/showyourwork/cli/commands/sync.py
+++ b/showyourwork/cli/commands/sync.py
@@ -1,0 +1,19 @@
+import os
+
+from ... import exceptions, logging, paths
+from ...config import edit_yaml
+from ...git import get_repo_branch
+from ...overleaf2 import Overleaf
+
+
+def overleaf_sync(force):
+    branch = get_repo_branch()
+    with edit_yaml("overleaf.yml") as config:
+        version = config["overleaf"].get("version", 1)
+        assert version == 2
+
+        url = config["overleaf"].get(
+            "url", f"https://git.overleaf.com/{config['overleaf']['id']}"
+        )
+        local_sha = config["overleaf"].get("local_sha", None)
+        remote_sha = config["overleaf"].get("remote_sha", None)
--- a/showyourwork/cli/commands/sync.py
+++ b/showyourwork/cli/commands/sync.py
@@ -1,6 +1,5 @@
-import os
 
-from ... import exceptions, logging, paths
+from ... import logging, paths
 from ...config import edit_yaml
 from ...git import get_repo_branch
 from ...overleaf2 import Overleaf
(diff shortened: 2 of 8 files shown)


MilesCranmer:
Just to check: you are planning to use merges, rather than rebases, right? I think rebases won't work on the overleaf side since you aren't allowed to force push to their hosted repos.

> How do we handle pushing changes back to Overleaf? I think it's nearly identical, but we'll need to think about handling merge conflicts. We could save a temporary directory with the checked out Overleaf repo.

The merge conflicts would be solved when merging to the SYW repo, right? If so I don't think fixing them would be necessary going back. By merge conflicts here, do you mean new merge conflicts due to other users editing the overleaf repo during the SYW build?

> Still need to think about how to handle migration from existing Overleaf integrations, and start up for new projects.

Since we are assuming the repos have distinct histories, I think this part actually won't be difficult. An existing overleaf project should look basically the same as a newly created one, right?

> What about the front end? I like the idea of adding a showyourwork sync command to execute this update, and perhaps an showyourwork build --sync command that would work like the current showyourwork build, syncing every time.

Those sound great to me!

MilesCranmer:
How will this handle auto-generated LaTeX snippets like tables (i.e., the `\variable{}` command)? I'm assuming those are to be treated basically the same as figures, right?

dfm:
Thanks @MilesCranmer!

> The merge conflicts would be solved when merging to the SYW repo, right? If so I don't think fixing them would be necessary going back. By merge conflicts here, do you mean new merge conflicts due to other users editing the overleaf repo during the SYW build?

Yes - this is exactly the issue I was referring to! I wasn't aware of the force push restriction, but I don't think there's a problem with that. The way I was imagining the workflow would only ever fast forward the temporary local copy of the Overleaf repo, which gets cloned at the beginning of the build. There would be an issue if we hit the race condition you mention when pushing back, and we just want to handle it gracefully. But, since that merge is always of common histories, we'd never need to rebase.

> Since we are assuming the repos have distinct histories, I think this part actually won't be difficult. An existing overleaf project should look basically the same as a newly created one, right?

Not quite! The way syw currently handles new repos is that it wipes them, or refuses to push anything if it thinks their may have been changes. I think that's a fine approach to keep for new repos, but it would be nice to have a sensible migration flow for projects that are currently using the existing integration scheme. Certainly doable, but will take some thought to make a reasonable interface.

> How will this handle auto-generated LaTeX snippets like tables (i.e., the \variable{} command)? I'm assuming those are to be treated basically the same as figures, right?

Yes! I haven't implemented that logic yet, but the idea is that, like in the current implementation, you would also have a set of glob patterns to push (and possibly pull, but probably not). Thinking about it more, syw should be able to generate that list automatically since it knows which files are generated and required by the manuscript. @rodluger: Is it straightforward to get that list (the paths to all the generated manuscript dependencies: figures, tables, snippets, etc) via the Python interface?

rodluger:
> @rodluger: Is it straightforward to get that list (the paths to all the generated manuscript dependencies: figures, tables, snippets, etc) via the Python interface?

You can get the list of all dependencies of the PDF generation step from

```python
deps = set(config["dag_dependencies"][config["ms_pdf"]])
```

where ``config = snakemake.workflow.config``.

If you need recursive dependencies (i.e., dependencies of those dependencies, which may not be explicit dependencies of the PDF generation step), you can instead do

```python
deps = set(config["dag_dependencies_recursive"][config["ms_pdf"]])
```

(I don't think that's the case here, but thought I'd mention it.)

Then, you can get all files that are programmatically generated by the workflow from

```python
outputs = set([file for job in dag.jobs for file in job.output])
```

where you'll need to hackily grab the workflow DAG from

```python
from showyourwork.patches import get_snakemake_variable
dag = get_snakemake_variable("dag")
```

What you want, then, is the intersection of these sets:

```python
generated_deps = deps & outputs
```

The easiest place to do this is in the [WORKFLOW_GRAPH](https://github.com/showyourwork/showyourwork/blob/6ca083ff1b33910043216adc0b20c80e8c6706a9/showyourwork/workflow/rules/dag.smk#L161) input function, which blocks the PDF generation rule until the DAG has been built.

*EDIT*: This approach will also output some hidden dependencies of the workflow, like the ``.showyourwork/flags/SYW__DAG`` flag. You could add a filter and select only files under ``src/tex``.

MilesCranmer:
@dfm let me know if you want a guinea pig to test this out! What's the current status of things?

dfm:
@MilesCranmer — Awesome and thanks! It's getting close and all the git logic more or less works, but it's not actually properly linked into the workflow yet. I'll ping you as soon as I have a working prototype!

PR #162, opened 2022-08-17
fix syw version resolution bug for relative paths

rodluger:
Fix bug related to relative showyourwork paths in the config file


dfm:
whoops - should probably have a test for that...

dfm:
I didn't add tests for paths since it checks to make sure they exist, but it wouldn't be hard to do it properly!

rodluger:
I think this is OK as is, unless we start messing with Windows path support, in which case unit tests might be nice.

PR #168, opened 2022-08-19
Reorganize test workflows a little

rodluger:
Separates unit tests, local integration tests, and remote integration tests into their own ``.yml`` workflow files to make it a bit easier for devs to understand what's going on.


dfm:
You'll need to update the README badge too, but otherwise LGTM!

rodluger:
Ah, right -- good catch.

PR #170, opened 2022-08-19
Remove provided classes

### LINKED ISSUE #154: Should we just remove the packaged classfiles?

rodluger:
As mentioned in #154.


rodluger:
Merging this into `staging` so we can run the remote integration tests.

PR #171, opened 2022-08-19
Remove provided LaTeX class files

### LINKED ISSUE #154: Should we just remove the packaged classfiles?

rodluger:
This is PR #170 reincarnated on the `staging` branch.

Closes #154.


rodluger:
Note that this PR will break current users' workflows if they rely on the ``aastex631``, ``mnras``, or ``aa`` built-in class files. However, users will only run into this issue if they upgrade the version of ``showyourwork`` specified in the ``showyourwork.yml`` config file of a given article repo. Running a workflow that requires an earlier version of ``showyourwork`` will still work regardless of whether or not the locally installed ``showyourwork`` includes the changes provided by this PR, since articles are always built in an isolated conda environment with the author-defined ``showyourwork`` version.

I'll add some notes about this to the docs, then I'll mark this ready for review.

PR #177, opened 2022-08-30
Fixing latex error for ARA&A class

dfm:
The most recent release (probably the switch to `marginpar`) broke the build for the annual reviews classfile. It would fail to build with a "lost float" error, when making the title page. This makes the abstract icons render inline in the abstract, instead of as a margin icon and everything seems to work: https://github.com/dfm/araa-gps

### DIFF
--- a/showyourwork/workflow/resources/styles/build.tex
+++ b/showyourwork/workflow/resources/styles/build.tex
@@ -129,15 +129,10 @@
 % Add a GitHub link to the end of the abstract
 \let\oldabstract\abstract
 \let\oldendabstract\endabstract
-\renewenvironment{abstract}{%
-  \oldabstract%
-}{%
-  % AASTex631 hack to reduce margin separation
-  % in the abstract of a two-column document
-  %\ifcsname if@two@col\endcsname%
-  %\if@two@col\setlength{\marginparsep}{-45pt}\fi%
-  %\fi%
-  \marginpar{%
+\@ifclassloaded{ar-1col}{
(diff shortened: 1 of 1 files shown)


rodluger:
Nice! If you have the energy, now might be a good point to create a unit test for this...

dfm:
Agreed! For now, I'd rather merge this and open a separate issue to track that since it won't be quick!

MilesCranmer:
Let me know what you think, but my sense is that hardcoding tweaks like [these](https://github.com/showyourwork/showyourwork/blob/3446303cd27b5730ba931e38fc7f194b5f07bbea/showyourwork/workflow/resources/styles/build.tex#L62-L110) for specific journals is a code smell–i.e., it suggests a deeper issue that needs to be addressed, rather than adding additional tweaks. This is especially true since we want people from other disciplines to use SYW! If it only works reliably for astronomy journals, then it will be discouraging for outside users.

The easiest test for this is: does it work for astronomy journals when we *don't* have such tweaks?

Is there a generic way to fix these types of issues, or maybe a way to make it easier for a user to tweak these things? e.g., maybe there could be better controls for users to specify the position of the icons? 

(arXiv can reliably add their text with the arXiv ID on top of rendered papers - maybe a similar strategy could be used?)

dfm:
Agreed in the long run! In the short term, I need a version that works with this class so 🤷‍♂️

MilesCranmer:
That is indeed a good reason 👍 😅

PR #183, opened 2022-09-12
Check if figure, abstract defined before redefining them

### LINKED ISSUE #178: Support for `documentclass`s which do not have `figure` or `abstract`
It seems like `showyourwork` doesn't support document classes which do not have the figure environment defined. For example, the following `ms.tex`:
```latex
\documentclass{letter}
\usepackage{showyourwork}
\begin{document}
Test
\end{document}
```
gives the error:
```
error: showyourwork.tex:46: LaTeX Error: Environment figure undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help
error: halted on potentially-recoverable error as specified

! LaTeX Error: Environment figure undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.46 \renewenvironment{figure}
                              {%
No pages of output.
```
You can get this to work with the following modification:
```latex
\usepackage{graphicx}
\usepackage{float}
\usepackage{caption}
\newfloat{figure}{htbp}{figs}
```
before the `showyourwork` import. 

I also see a similar error when `abstract` is undefined:
```
error: showyourwork.tex:132: LaTeX Error: Environment abstract undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help
error: halted on potentially-recoverable error as specified

! LaTeX Error: Environment abstract undefined.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.132 \renewenvironment{abstract}
                                 {%
No pages of output.
```
This can be fixed with:
```latex
\newenvironment{abstract}{}{}
```

I think it is preferable if `showyourwork` could be used to render *any* valid LaTeX document, albeit with fewer features than when normal document classes are used. Therefore I think the parts of `showyourwork.tex` which rely on `figure` and `abstract` should be wrapped inside a `try-catch` (whatever that is in LaTeX). What do you think @rodluger?

For reference: I'm trying to port my CV to showyourwork so I can programmatically generate the LaTeX, but I'm running into quite a few issues since it uses a non-standard document class.

rodluger:
Fixes #178 so we can use `showyourwork` with (potentially) any documentclass.

### DIFF
--- a/showyourwork/workflow/resources/styles/build.tex
+++ b/showyourwork/workflow/resources/styles/build.tex
@@ -12,6 +12,7 @@
 \RequirePackage{xspace} % better spacing after icons
 \RequirePackage{graphicx} % includegraphics
 \RequirePackage{stackengine} % stacking figure links
+\RequirePackage{etoolbox} % ifcsmacro
 
 % Is this running on GitHub Actions?
 \newif\ifOnGithubActions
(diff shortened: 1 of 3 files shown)


MilesCranmer:
How do I test this locally? I tried installing it, but not sure what to put for `showyourwork.yml` - it complains it can't find the version.

MilesCranmer:
Ah, nevermind - I found the `version.path` argument!

MilesCranmer (approved):
Reviewed the code, and tested locally and it's working 👍

PR #184, opened 2022-09-12
Watermark version of showyourwork logo and GitHub Icons

### LINKED ISSUE #181: Implement generic showyourwork logo & abstract icons for any journal
This is an issue raised by @MilesCranmer in https://github.com/showyourwork/showyourwork/issues/177#issuecomment-1236261572:

> Let me know what you think, but my sense is that hardcoding tweaks like [these](https://github.com/showyourwork/showyourwork/blob/3446303cd27b5730ba931e38fc7f194b5f07bbea/showyourwork/workflow/resources/styles/build.tex#L62-L110) for specific journals is a code smell–i.e., it suggests a deeper issue that needs to be addressed, rather than adding additional tweaks. This is especially true since we want people from other disciplines to use SYW! If it only works reliably for astronomy journals, then it will be discouraging for outside users.
>
> The easiest test for this is: does it work for astronomy journals when we *don't* have such tweaks?
>
> Is there a generic way to fix these types of issues, or maybe a way to make it easier for a user to tweak these things? e.g., maybe there could be better controls for users to specify the position of the icons? 
>
> (arXiv can reliably add their text with the arXiv ID on top of rendered papers - maybe a similar strategy could be used?)

Currently, the strategy is to tweak the page header and abstract layout on a per-journal basis in the `showyourwork` stylesheet, which takes a lot of work, is not future-proof, and won't be sustainable if we hope to target other disciplines. As @MilesCranmer suggested, perhaps the simplest approach is to add an overlay/watermark in the margin like arXiv does. This watermark should include both the showyourwork logo and the GitHub / GitHub Actions hyperlink icons.

### LINKED ISSUE #182: Small "1" appearing in left margin of abstract
image

This behavior is new (I think) as of ~`0.3.1`, but I haven't been able to track the commit at which it started happening yet. See [here](https://github.com/showyourwork/test-default/blob/main-pdf/ms.pdf) for the pdf generated by the latest build of the `test-default` repo (warning: not a permalink!).

@MilesCranmer @dfm any idea what's causing this?

rodluger:
This PR will generalize the `showyourwork` logo and abstract icons so that they can work with any documentclass.
Fixes #181, fixes #182.


rodluger:
OK, this is mostly ready for other people's eyes (@MilesCranmer @dfm).

The new implementation uses `tikz` to draw the three icons you see in this image at the top left of the first page of the document (the last icon is the "reproducibility status" icon, which becomes a green checkmark when you build on GitHub Actions).

image

Users can control some aspects of the watermark in the config file, e.g.,:

```yaml
watermark:
  enabled: true
  xpos: 0.10
  ypos: -0.20
  angle: 0.0
```

(position in inches relative to top left of page; angle in degrees).

The image above shows the default -- happy to chat about different choices (i.e., top right, or rotated in the right margin, etc.).

This is _mostly_ documentclass-independent, except that for some reason `tikz` cannot properly locate the left edge of the paper in `aastex631`: it believes the left edge is about `0.65in` to the left of where it actually is on the PDF (likewise for the right edge). There is probably some weird `geometry` hack in `aastex631.cls` that's causing this, but it wasn't obvious how to work around this. So @MilesCranmer may still be unhappy with me because I need some aastex-specific code in the showyourwork stylesheet [here](https://github.com/showyourwork/showyourwork/pull/184/files#diff-41cb79f656d48e00c02a7cf790d2b28df7e1a77480d65e684eaa485146cb0b76R76). However, this implementation works out of the box for `aa`, `mnras`, and `letter`. It really is just something peculiar about `aastex`.

Note that I looked into how arXiv does its watermark and it's not a LaTeX thing -- it adds the watermark directly to the PDF, after it is rendered. We could easily implement something like that in `showyourwork` in a postprocessing step, but I don't like this solution since articles submitted to the arXiv will not display the showyourwork logo (since we can't control the build step on arXiv).

MilesCranmer:
Awesome! Nice job. I think this strategy is the best way forward, even though there is some class-specific code.  I think it's completely fine if there are packed defaults for certain classes, so long as it's easy for a user to adjust them without hacking the LaTeX template. This gets a :+1: from me.

I would also add that example config to the automatically-generated one, so that it's more obvious how it can be adjusted without needing to delve into the docs $^{[1]}$.

Regarding the look of it, I am not a huge fan of the `/` since my brain parses it as being related to hierarchies - i.e., `/showyourwork/github/status`. Maybe just `|`?

> Note that I looked into how arXiv does its watermark and it's not a LaTeX thing -- it adds the watermark directly to the PDF, after it is rendered. 

Very interesting, I didn't know this. Your `tikz` approach seems like a good solution; I agree with your points about avoiding PDF post-processing.

---

$[1]$ Actually, maybe this could be done for other (stable) config keys as well? Much like how sphinx initializes a config with all options listed, but commented out, or their defaults listed. (Users are sometimes bad at reading docs 😅 so I think making it obvious how to adjust things is always good.)

MilesCranmer:
Also - I think the "show your work!" might be too noticeable for my taste, especially since it is bright red and a different font. Is there a way it could be changed to "syw!" or maybe "built with show your work!" and maybe in grayscale?

rodluger:
Cool. I can start a new PR to add the various config options to the cookiecutter template -- I think that's a good idea.

I just played around with a bunch of alternative placements / sizes / colors of the logo and wasn't happy with any of them. I'd like to keep it red and spelled out in full for branding. But we could add some transparency to it, like arXiv does with their watermark. I'm less picky about the GitHub logos -- we could change the separators as you suggested, or spell stuff out in full, like "built from rodluger/repo_name using showyourwork".

But, overall, I think it does look weird to have this stuff floating off in a corner. It looked better when it was worked into the page heading, but of course that's super class-specific. My only other thought is to mimic an actual rubber stamp effect, which could look more natural sitting in the top (right?) corner. I'll try to come up with something that doesn't look cheesy and post it here.

rodluger:
image

Too cheesy?

dfm:
I legit lol-ed

MilesCranmer:
I actually love this.

It’s less distracting now, since the red is a lighter shade, and is simultaneously more aesthetically pleasing! I also really like it from a design standpoint as the visual matches the meaning: articles which don’t have this stamp are ultimately non-reproducible. The stamp gives me more confidence in a paper. It makes for great branding too. 

My only comment would be the details of the stamp might be overly stylized so it appears slightly in jest rather than serious. Maybe this is due to the diagonal alignment of the text? In other words, you might have done too good of a job making it look like a stamp :) maybe it’s good though!

rodluger:
Ha, so one "yes awesome" and one "definitely not". The other thing I considered was a page curl effect, like this one:

image

We can make it smaller too, and/or add additional icons. The thing about the `showyourwork!` logo is that its aspect ratio is too long (even if we shorten it to just `syw!`). I've been meaning to create a more compact one a la the GitHub logo, so perhaps now is the time.

dfm:
I don't know that `"lol" == "definitely not"`... I think that all of these are awesome!!

MilesCranmer:
I like the curled paper one too. Although, if I were to see this in a paper (with the GitHub icon only), I would just assume it means the code is open-sourced, not that the paper is as reproducible as a SYW-built paper. With a SYW icon it might be a good option though! It's even more aesthetically pleasing.

I also think that explicitly stating "reproducible article" is a nice touch though. If I don't know what SYW is and I see the logo, I might just guess it is some science collaboration framework or workflow, whereas including a 2-word description tells me what it is.

rodluger:
Ok, let's go with the stamp for now, since it (1) will print better (2) allows us to say "reproducible article" and (3) is much easier to generalize across documentclasses with finicky page geometries. Here's what it currently looks like (as of the most recent commit to this PR):

image

and here's what the user can customize:

```yaml
  stamp:
    enabled: true
    xpos: 0.50 # in inches from the right edge
    ypos: 0.50 # in inches from the top edge
    size: 0.75 # in inches (width of image)
    angle: -20 # in degrees
```

Clicking on the stamp takes the user to the GitHub page for `showyourwork`. It would be nice, however, to also link to the paper repository. Previously we did this via a GitHub icon, but I've always wanted to include the `user/repo` slug in text somehow to make it immediately obvious to the reader. I'm not sure how/where to include this information with the stamp implementation -- any thoughts?

(Perhaps we can instead link to the paper repo from the stamp?)

MilesCranmer:
Looks good! I like that you can click it.

I have no idea how hard it would be, but it would be **amazing** if the slug could be included in the stamp somehow- ideally, right below... with the same orientation and color. Then, I would put a little GitHub icon (also in red stamp-style) right next to the slug.

MilesCranmer:
i.e., the slug would be generated via tikz, overlaid right below the stamp - maybe with some transparency that makes it look like the static stamp?

MilesCranmer:
Oh, and I would definitely make the stamp link go to the GitHub page for the paper, rather than the `showyourwork` repo. They'll be able to find `showyourwork` either way if they want to.

rodluger:
> i.e., the slug would be generated via tikz, overlaid right below the stamp - maybe with some transparency that makes it look like the static stamp?

Haha let's see what I can do.

rodluger:
How's this?

image

```yaml
stamp:
    format: \bfseries\sffamily\normalsize
```

Currently, the orientation of the slug is independent of the rotation of the stamp, but I could rotate it by the same angle. What do you think @MilesCranmer ?

MilesCranmer:
Looks good! +1 for rotating by the same angle. Is there also a way to include a red GitHub icon so it's clear what the slug is for?

Also - have you tried making the text in a line, rather than wrapping around the stamp? I am just wondering how that looks instead... I'm not sure about the look of the `/` in curved text.

rodluger:
Rotated:

image

As for including a GitHub icon: the path transform I'm using in TikZ does not seem to support `fontawesome` symbols, so I haven't been able to figure out how to prepend the GitHub logo to the slug (I was thinking `logo / slug` or `logo: slug`). I can place the logo at the top, but it looks kinda weird.

As for text in a line -- I didn't like the look, especially now that we're rotating it with the stamp.

MilesCranmer:
How about putting the full URL (i.e., `github.com/rodluger/test`)? We all spend a lot of time on GitHub so it's immediately obvious what the `user/repo` string is, but for others it might be unclear.

rodluger:
Sure! Done:

image

MilesCranmer:
Awesome!

MilesCranmer:
Okay, just tried it out, using https://github.com/kazewong/SymbolicGWPopulation_paper as a test.

I am currently seeing this issue in the tectonic log:

```

Package hyperref Warning: Rerun to get /PageLabels entry.

ABD: EveryShipout initializing macros
File: showyourwork-stamp.pdf Graphic file (type pdf)
<use showyourwork-stamp.pdf>


LaTeX Font Warning: Font shape `TU/ptm/b/n' undefined
(Font)              using `TU/ptm/m/n' instead on input line 68.

LaTeX Font Info:    Trying to load font information for TU+phv on input line 68
.
LaTeX Font Info:    No file TUphv.fd. on input line 68.

LaTeX Font Warning: Font shape `TU/phv/b/n' undefined
(Font)              using `TU/lmr/m/n' instead on input line 68.

! Missing $ inserted.
<inserted text> 
                $
l.68 \begin{document}
                     
No pages of output.
```
Any idea what this is?

MilesCranmer:
Hm, it still gives me this error when I remove all packages and text from the document... Very strange

MilesCranmer:
I tested a few of the commits - it seems like 6a7d13d96b1649497ae89c3454841335f251b7cc is where this error was introduced.

rodluger:
Ha, it's because there's an underscore in the repo name! Good catch! I can fix this.

rodluger:
Fricking TeX.

showyourwork/config.py
```diff
@@ -401,6 +398,17 @@ def parse_config():
     # Git info for the repo
     config["git_sha"] = git.get_repo_sha()
     config["git_url"] = git.get_repo_url()
+    config["git_short_url"] = (
+        config["git_url"]
+        .replace("https://", "")
+        .replace("http://", "")
+        .replace("_", "-")
```
MilesCranmer: Would it be better to stick it inside a `\url{}` instead of changing to `-`?

rodluger: Definitely, or any of `\_`, `$_$`, ``\char`_``... But I'm using a "path decoration" in TikZ (see [here](https://github.com/rodluger/showyourwork-dev/blob/watermark/showyourwork/workflow/resources/styles/build.tex#L88-L95)), which appears to only allow plain text with very limited formatting and no special characters. There's no error if you include a special character, but the build just hangs indefinitely. I think this is probably a bug in TikZ, but I haven't been able to track anything down online. This is also, by the way, the reason I wasn't able to include the GitHub logo along the path.

rodluger: The best approach here would probably be to abandon this "path decoration" approach entirely, but I don't know how else to generate curved text like that. There may be a different way of doing it with TikZ, but I'm not sure.

MilesCranmer: Ah, I see, thanks! That is definitely tricky since `-` and `_` will map to different repositories in GitHub URLs... not sure what the best strategy is or if there is any way to get a similar character working.

rodluger: Ha, the URL in the hyperlink itself is fine -- that doesn't get changed, and the link points to the correct repo. It just _displays_ as a hyphen. I hate it, but I currently don't have an alternative.

rodluger: Ah! [This](https://tex.stackexchange.com/a/138730) is why that's happening: TikZ places one character at a time along the path, so things like `$`, `\`, etc., which are parts of multi-character expressions, cause it to fail. Let me look into this more -- seems like that thread had a solution.

rodluger: Looks like we just need to do wrap the thing in curly braces. I'll work on it after lunch!

rodluger: OK, trick was to replace "_" with "{\_}". Working now!

rodluger:
OK, I figured out how to include the github logo. I also added a max length so the text gets truncated after 50 characters:

image image image

MilesCranmer:
Nice! Thanks!

However now it seems to display the logo on a separate page: 
Screen Shot 2022-09-14 at 3 48 39 PM

Not sure if this is something with the journal class or not...

MilesCranmer:
Yep - when I turn off the journal class, this new page goes away. Maybe the journal class is running `\clearpage` or something before creating the title. I wonder if generating the logo when `\maketitle` is called would fix this?

rodluger:
Ha, funny! Yeah currently it renders the logo during `\AtBeginDocument`. We could override `\maketitle` instead -- let me work on that.

MilesCranmer:
I wonder if there is a way in Tikz to "move the cursor" to a certain page - after it has already been generated - and add graphics after-the-fact? I am worried hooking into `\maketitle` might introduce some issues

This looks like another hack that might work: https://tex.stackexchange.com/a/140329/140440, although deleting the first page might also delete the logo...

rodluger:
Oh, looks like I was too quick in implementing your requested change! The latest commit overrides `maketitle` and seems to work for me. Can you test it?

I can't imagine what issues this might introduce -- we're basically just appending some code to the end of the `maketitle` command.

MilesCranmer:
Hm, now the stamp doesn't render at all. Very strange.

rodluger:
> I can't imagine what issues this might introduce

Well you proved me wrong pretty quickly! It works perfectly for `aastex631` (see [here](https://github.com/showyourwork/test-default/blob/main-pdf/ms.pdf)).

I'll poke around your build and see what I can find.

rodluger:
Could it be that `icml2022` simply doesn't use `\maketitle`? It looks like the stylesheet is just manually printing text with some vertical bars at the top of the first page: https://github.com/kazewong/SymbolicGWPopulation_paper/blob/main/src/tex/icml2022.sty#L403

MilesCranmer:
Ah, yeah! Oops.

rodluger:
So we're back to the problem of not having a good solution that generalizes to all possible stylesheets :/
The advantage of `\AtBeginDocument` is that it's always executed, so my guess is that's our best bet, as long as we can circumvent the issue you raised earlier with the extra blank page. I think deleting that page isn't the right way to go -- can we try to figure out exactly why it's generating a new page?

rodluger:
Perhaps the simplest (but not necessarily best) solution is to define a command like `\stamp` and stick it in the default `ms.tex`. Then users can move it around as needed?

rodluger:
But hang on -- I think I understand the issue with `icml`. They define all the title stuff in their own `\AtBeginDocument`, so we just need a way to ensure that our `\AtBeginDocument` hook gets executed _after_ the one in the `icml` stylesheet. There may be a way to do this (see [here](https://tex.stackexchange.com/a/602677)) but I'm not sure yet.

MilesCranmer:
`\stamp` seems like an elegant approach in case the `\AtBeginDocument` doesn't work out. You could even pass your angle/size information there, rather than in the config file. 

The only thing is: knowing how I typically use LaTeX packages, I would never know the `\stamp` command exists unless there is a warning or error message about it... Maybe the `showyourwork` build could print out a warning if it can't find the `\stamp` command - and a user could disable this warning by explicitly disabling the stamp in the config file (i.e., make the desired behaviour - having the stamp - the easier one for the user).

rodluger:
OK, let's give this one more try -- @MilesCranmer can you build using the latest commit? I switched from `\AtBeginDocument` to `\AfterEndPreamble`, which [SO](https://tex.stackexchange.com/a/565208) suggests always executes _after_ `\AtBeginDocument`.

MilesCranmer:
Darn, it's back to printing the logo on a separate page before the title.

MilesCranmer:
If it helps, it seems like it's printing after the header line:
Screen Shot 2022-09-14 at 4 54 41 PM
but before everything else.

rodluger:
We could explore some of the other hooks, like `\AddToHook{shipout/background}`, which I think gets called after each page is rendered. Then we'd have to do some bookkeeping to ensure this only happens for the first page. Perhaps something like [this](https://tex.stackexchange.com/a/621695). I can play around with this tomorrow. Or, if you're interested, feel free to play around with different hooks [here](https://github.com/rodluger/showyourwork-dev/blob/watermark/showyourwork/workflow/resources/styles/build.tex#L76).

rodluger:
@MilesCranmer My last attempt of the day: using `\AddToShipoutPictureBG` instead of `\AtBeginDocument`. Can you test now? If it doesn't work, I'll fork your repo and debug locally tomorrow. Thanks!

rodluger:
Aha!
image

MilesCranmer:
Amazing!!! It works 🎉  Nice job getting that done!

showyourwork/workflow/scripts/pdf.py
```diff
@@ -23,9 +23,22 @@
...
+    ((* if stamp.enabled *))
+    \AddStamptrue
+    ((* else *))
+    \AddStampfalse
+    ((* endif *))
+    \newcommand{\syw@stampX}{((- stamp.xpos -))}
+    \newcommand{\syw@stampY}{((- stamp.ypos -))}
+    \newcommand{\syw@stampSize}{((- stamp.size -))}
+    \newcommand{\syw@stampAngle}{((- stamp.angle -))}
+    \newcommand{\syw@stampFontSize}{\footnotesize}
```
MilesCranmer: Could this be a configurable too? Making the stamp size smaller causes the text to wrap all the way around, otherwise.

showyourwork/config.py
```diff
@@ -313,6 +313,17 @@ def parse_config():
...
         config["optimize_caching"] = config.get("optimize_caching", False)
 
+        #: Showyourwork stamp settings
+        config["stamp"] = as_dict(config.get("stamp", {}))
+        config["stamp"]["enabled"] = config["stamp"].get("enabled", True)
+        config["stamp"]["size"] = max(0.75, config["stamp"].get("size", 0.75))
+        config["stamp"]["xpos"] = config["stamp"].get("xpos", 0.50)
+        config["stamp"]["ypos"] = config["stamp"].get("ypos", 0.50)
+        config["stamp"]["angle"] = config["stamp"].get("angle", -20)
+        config["stamp"]["max_url_len"] = int(
```
MilesCranmer: If fontsize is configurable, then `max_url_len` should be configurable as well. (Would be too complicated to infer from the fontsize...)

showyourwork/workflow/resources/styles/build.tex
```diff
@@ -43,133 +52,73 @@
...
-    \oldjournalinfo%
-    \vskip6\p@{\jinfofont%
-      Compiled using\begin{minipage}{7.3em}\includegraphics[width=7.3em]{showyourwork-logo.pdf}\end{minipage}\endgraf%
-      \ifcsname gitHeader\endcsname
-        \,(\gitHeader)\endgraf
-      \fi
+  % Define the stamp
+  \newcommand{\syw@stamp}{%
+    \pgfmathsetmacro{\radius}{\syw@stampSize*0.66}%
+    \pgfmathsetmacro{\sx}{\syw@stampX + \syw@stampXOffset + \syw@stampSize/2}%
```
MilesCranmer: Just to double check - if `\radius` is 0.66 the stamp size, then why is `\sx` only offset by 0.5 the stamp size (rather than 0.66*0.5)?

showyourwork/workflow/scripts/pdf.py
MilesCranmer: (i.e., the user could specify `"\footnotesize"`)

showyourwork/workflow/resources/styles/build.tex
```diff
@@ -43,133 +52,73 @@
...
+            text along path,
+            text={|\bfseries\ttfamily\syw@stampFontSize|\syw@shortUrl},
+            text align=center,
+            text color=sywStampRed
+          }]
+          ($(STAMP.center) - (\ax in, \ay in)$)
+          arc
+          [start angle={\syw@stampAngle+90},
+          end angle={\syw@stampAngle+450},
+          radius=\radius in];
```
MilesCranmer: Would love this (radius of the text) to be a configurable as well - smaller stamp sizes would look better with text that is farther out.

showyourwork/workflow/resources/styles/build.tex
```diff
@@ -43,133 +52,73 @@
...
+      \end{tikzpicture}%
```
MilesCranmer: I also wonder if you'd like to put the SYW version information somewhere in the stamp. e.g., perhaps in the future, the stamp style changes. I wouldn't want people to think I am using a knockoff SYW, trying to pretend my work is reproducible :P - showing the version number would explain an older stamp style.

(Maybe this is too much info/too much work - fine by me!)

MilesCranmer (commented):
Looking good! Here are some brief suggestions.

MilesCranmer:
One other (hopefully last!) comment - a tiny nitpick: if possible, could you use a vector graphics version of the stamp, rather than rasterized? No worries if not possible!

rodluger:
> One other (hopefully last!) comment - a tiny nitpick: if possible, could you use a vector graphics version of the stamp, rather than rasterized? No worries if not possible!

Not easily (with the tools I have), but I replaced it with a higher DPI. Does that work?

showyourwork/workflow/resources/styles/build.tex
rodluger: `\radius` here is the radius of the text arc, which is `4/3` times the radius of the stamp, or `2/3` its diameter (stored in `\syw@stampSize`). I think it's correct as is.

showyourwork/workflow/scripts/pdf.py
rodluger: My thought was to keep the font size fixed and use a monospace font, so I know exactly how long to truncate the URL to for any stamp size. The current implementation should work out of the box for any value of the stamp size, but you're right that it's not very tunable.

The opposite approach, which I think is what you're suggesting, is to make everything customizable. That's fine by me, but certain combinations of parameters will lead to horrible-looking stamps.

One alternative is to fix the font size relative to the size of the stamp. We don't have to do this programmatically -- we could just render it as it currently appears for the default stamp size, but then have ``stamp.size`` scale _everything_ up (including the text). Should be easy with a `\resizebox` after the TikZ graphic is rendered. That way we can ensure the stamp always looks good and keep the branding consistent across papers. What do you think?

showyourwork/config.py
rodluger: See comment above. But even with my suggested approach, I'd be ok with letting the user change `max_url_len`, as long as the default is sensible.

showyourwork/workflow/resources/styles/build.tex
rodluger: I think my suggestion above would resolve this point as well.

showyourwork/workflow/resources/styles/build.tex
rodluger: This is a good idea. It's not too much work -- should be easy with TikZ. The only issue is that the version can be a number, a tag, a branch, a commit, a path, or a fork. Some of those won't fit in the stamp. Perhaps we could display the version as long as it's a pip-installable version, and (maybe) just ``dev`` otherwise?

showyourwork/workflow/scripts/pdf.py
MilesCranmer: The reason I asked about this is because I tried making the stamp smaller, but it starts overlapping, since the text doesn't scale:
Screen Shot 2022-09-15 at 12 13 23 PM
Maybe the `max_url_len` needs to scale differently due to the radius of the text being different at the top of characters vs bottom of characters?

But in any case, I think either configurable fontsize would help this, or your suggestion of a `\resizebox` would probably be even better! An overall `stamp.size` would be awesome and the best way.

showyourwork/workflow/resources/styles/build.tex
MilesCranmer: Sounds good to me!

MilesCranmer:
Yep, a higher DPI one is good enough for me!

showyourwork/workflow/scripts/pdf.py
rodluger: Resizebox version (actually scalebox) implemented.

showyourwork/workflow/resources/styles/build.tex
rodluger: Working on this right now.

rodluger: Done. The version now appears inside the stamp if it's a pip-installable version; otherwise it displays ``dev``.

rodluger:
Current layout of the stamp:

image &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; image

Looks like I misspelled my own name, but oh well.

rodluger:
Different values for the stamp size with @MilesCranmer's icml paper (default first):

image

image

image

rodluger:
@MilesCranmer This is ready for review again. The one thing that's weird is that for some reason commands like `\bfseries` and `\ttfamily` (and basically every other font transform) don't work with your manuscript. Perhaps `icml2022` uses a font that doesn't have alternative styles? In any case, that currently causes the text in the stamp to use the standard serif font rather than the custom monospace font. Not sure how to fix this.

rodluger:
I also want to give a chance for other contributors to give input on this. @dfm @katiebreivik @mathren @adrn @jfcrenshaw any thoughts before we merge this?

MilesCranmer:
> The one thing that's weird is that for some reason commands like \bfseries and \ttfamily (and basically every other font transform) don't work with your manuscript.

Strange. The only thing I could think of is the [line](https://github.com/kazewong/SymbolicGWPopulation_paper/blob/98770388f4ab6f0a89d6b2b95dab416f5d6fa17b/src/tex/icml2022.sty#L83):
```latex
\RequirePackage{times}
```
nothing else looks to tweak the font family. 

I tried using:
```latex
\usepackage[utf8]{inputenc}
\usepackage[T1]{fontenc}
```
before any package imports and this seems to fix it!

docs/config.rst
```diff
@@ -840,3 +840,130 @@ the |showyourwork| repo specified by ``version.fork``.
 
   version:
     ref: main
+
+
+.. _config.stamp:
```
MilesCranmer, line 845: I think `maxlen` is missing here?

showyourwork/config.py
```diff
@@ -313,6 +367,15 @@ def parse_config():
...
         config["optimize_caching"] = config.get("optimize_caching", False)
 
+        #: Showyourwork stamp settings
+        config["stamp"] = as_dict(config.get("stamp", {}))
+        config["stamp"]["enabled"] = config["stamp"].get("enabled", True)
+        config["stamp"]["size"] = config["stamp"].get("size", 0.75)
+        config["stamp"]["xpos"] = config["stamp"].get("xpos", 1.0)
+        config["stamp"]["ypos"] = config["stamp"].get("ypos", 1.0)
+        config["stamp"]["angle"] = config["stamp"].get("angle", -20)
+        config["stamp"]["maxlen"] = config["stamp"].get("maxlen", 30)
```
MilesCranmer: Could the default maxlen be increased to 40? For me personally I would rather have the text wrap around a little bit more before it displays a `...`.

MilesCranmer: (My sense is that many paper slugs are longer than 30 characters - although I don't have actual stats on this...)

adrn:
Hah, this is a cute idea. I don't have any strong feelings, but IDK if this is the right place to put a link to github: a) in the long term, github links are probably not something to rely on for truly reproducible work because it is a commercial enterprise that doesn't have the same commitment to archiving (vs. zenodo, software heritage, etc. - see: [decisions made by gitlab](https://docs.gitlab.com/ee/administration/inactive_project_deletion.html)), b) I think it looks ugly for long names to curve around the stamp, but that's aesthetic and subjective 😄.

MilesCranmer:
> in the long term, github links are probably not something to rely on for truly reproducible work because it is a commercial enterprise that doesn't have the same commitment to archiving (vs. zenodo, software heritage, etc. - see: [decisions made by gitlab](https://docs.gitlab.com/ee/administration/inactive_project_deletion.html))

Good point. Although this is a more general point than just this PR! I don't know how software heritage works, but luckily it looks like it automatically archives public GitHub repositories over time on some schedule.

Maybe one strategy could be to have a showyourwork GH action that automatically (with some regular cronjob) either:
1. Trigger a software heritage save request for the repository: https://archive.softwareheritage.org/save/, or
2. Push the source code to a zenodo archive?

If so, perhaps one of those could be linked to instead of the GitHub?

Then, if one day GH decides to delete old repos, at least everything is archived in long-term storage.

---

By the way @adrn - it looks like that GitLab setting is just for self-managed GitLab repos? i.e., not affecting gitlab.com itself. I would be very surprised if that setting was in place for public gitlab repos and also opt-out... Even [Google Code](https://code.google.com/archive/), which stopped in 2016, still has a publicly-available archive of all their code.

mathren:
This looks good (and I could see myself using it to hammer a point about reproducibility in a talk for example).
However, it also looks like something that will make it harder to deal with the journals (given their already reluctance to incorporate icons at figure margins). This is not to mean that I am against implementing this (especially if it can be toggled on/off without too much hacking).

Regarding this:
>a) in the long term, github links are probably not something to rely on for truly reproducible work because it is a commercial enterprise that doesn't have the same commitment to archiving (vs. zenodo, software heritage, etc. - see: [decisions made by gitlab](https://docs.gitlab.com/ee/administration/inactive_project_deletion.html))

I fully agree, and maybe linking the zenodo doi is a better strategy short term. Beyond this specific PR, it would be nice to be able to have a showyourwork pipeline independent on github actions, and that can run with multiple continuous integration tools. I also think different users may have different needs for long term hosting their data and code. Also getting too opinionated about zenodo could be a limitation in the future.

rodluger:
All good points, but I think they're all beyond the scope of this PR. `showyourwork` currently links to GitHub all over the place -- from the logo to the old abstract icons (which this PR replaces) to the icons in the figure captions. This PR simply changes the aesthetic of the abstract link, generalizing it to a watermark/stamp so it can work with any documentclass, including, e.g., ones without abstracts.

As for explicitly printing the URL of the repository in the stamp, we can certainly reconsider that. Point taken, @adrn, that it looks ugly. I think @MilesCranmer also didn't like the circular text. Our options then are to (a) find a different layout or (b) not print the URL (but still link to it). If anyone wants to take a stab at (a) please feel free; but my two cents on this are that it will be difficult to find a different layout that works well for long and short URLs alike. As for (b) perhaps that could solve the problem of link rot, at least in the printed version of the paper. Thoughts?

I started a separate issue where we can discuss reproducibility in the long term (#191). But regarding the comments above, I just want to say that `showyourwork` is explicitly designed, for better or for worse, with short-term reproducibility in mind. Several of the design decisions -- from integrating articles with GitHub repos to using `pip` and `conda` environments to manage dependencies -- are bound to cause trouble down the line as the web, the software, and the hardware we use evolve. Archiving the source code on Zenodo or Software Heritage won't be of much use unless typing things like `conda install -c conda-forge qt=5.12` and `pip install urllib3==1.9.1` still mean the exact same thing twenty years from now.

Instead, I think one of the main draws of `showyourwork` is that someone reading an article can very easily access its source code _today_: reviewers can easily inspect the source code of specific figures, and students or other researchers can easily clone the repository and build on it to write their next paper. I think making it easier for researchers today to do these things -- by fully exploiting the current dependency of the workflow on GitHub -- is worth the risk of broken links down the line.

The other thing to keep in mind, regarding a point @mathren brought up, is that `showyourwork` does not play well with many astronomy journals. There's a lot of reluctance/inertia on the editors' parts to include these links, for many of the reasons @adrn and others brought up. And they're not wrong -- `showyourwork` is not the perfect solution. But my hope is that if enough people adopt `showyourwork`, and enough editors need to go through the pain of deciding what to do about these potentially expiring links on a daily basis, then eventually the journals are going to have to do something about it. The point here being that I don't think we actually need to solve the problems related to long-term reproducibility -- we just need to get the conversation started.

katiebreivik:
I agree with @adrn on the long text wrapping -- and am in favor of removing the link text but using the image as an icon link to the repo. 

I also agree with @rodluger re:long-term reproducibility. I think that we shouldn't let perfect get in the way of continued conversation here. It's 100% worth continued work, but probably not for this particular PR.

adrn:
Agreed - let's discuss in the other issue, but I do support dropping the link text from the stamp (and leaving the icon as a hyperlink)!

rodluger:
May I suggest a compromise: allow users to display their GitHub URL in this fashion, but disable it in the config by default?

docs/config.rst
rodluger: Good catch. I'll add it.

showyourwork/config.py
rodluger: 40 is fine. I'll change it.

rodluger:
@MilesCranmer Any last comments, or are we ready to merge?

MilesCranmer (commented):
Looks good!

MilesCranmer (approved):

PR #189, opened 2022-09-14
Add option to not commit overleaf changes

### LINKED ISSUE #187: An invalid git hash is frequently saved in last_commit_sha.txt
On the workflow run for my [dfm/araa-gps](https://github.com/dfm/araa-gps) project the article cache restore seems to always fail because an invalid hash is being saved to `last_commit_sha.txt`. This causes the cache restore to fail, and means that the figure scripts never get run without re-setting the cache. I assume that the cause is that we're dirtying the repo somehow during the build and I'll try to track that down, but it would probably be good to handle this case. Some things that perhaps should happen:

1. If the cache restore / mod time update fails, we should probably _remove_ the cache and re-run the scripts.
2. We probably want to write `last_commit_sha.txt` _before_ running the build.

rodluger:
We don't need to commit the changes to the article repo after pulling from Overleaf on GitHub Actions, since we're not pushing back to the GitHub remote. This also helps with a bug @dfm found in the Actions caching step (see #187).

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -468,6 +468,27 @@ and make sure to check out :doc:`overleaf`.
             - src/tex/bib.bib
 
 
+.. _config.overleaf.gh_actions_sync:
+
+``overleaf.gh_actions_sync``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**Type:** ``bool``
+
+**Description:** Commit and push Overleaf changes to the GitHub remote when running on GitHub Actions?
+
+**Default:** ``true``
+
+**Required:** no
+
+**Example:**
+
+.. code-block:: yaml
+
+    overleaf:
+        gh_actions_sync: true
+
+
 .. _config.overleaf.id:
 
 ``overleaf.id``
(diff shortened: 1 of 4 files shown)


rodluger:
Fixes #187

dfm:
I think I might actually prefer to default to having the workflow push. If not, the Overleaf and syw versions of the manuscript will continue to diverge, until you run a local build, which doesn't seem ideal.

rodluger:
OK, we can make that an option under `overleaf:` in the config, and set the default to `true`. Can you confirm that the `git push` did the trick for you in your local fix?

rodluger:
@dfm I added the `overleaf: gh_actions_sync:` option, which defaults to `true`. Let me know what you think and if this solves your issue.

dfm:
I can confirm that the push fixed the issue on our project. I probably won't be able to try this implementation out until later this week, but I like the approach!

rodluger:
Fixes #187

rodluger:
Pinging @dfm for a review when you get a chance. Thanks!

dfm:
Sorry about the delay! This seems sensible to me. It's hard to test this in detail though since we mainly hit this when actively editing the Overleaf while also pushing to the repo. I'd say let's merge and I don't expect it will cause any real issues.

dfm (approved):

PR #196, opened 2022-09-21
Margin links for variables

rodluger:
Adds margin links wherever a `\variable` command is found in the manuscript. The links point to the line in the Snakefile (or other Snakemake file) where the rule that generates the programmatic output is defined (if the workflow can figure it out!)

### DIFF
--- a/showyourwork/workflow/resources/styles/build.tex
+++ b/showyourwork/workflow/resources/styles/build.tex
@@ -179,7 +179,16 @@
   }%
 }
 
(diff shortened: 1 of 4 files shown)


rodluger:
An example from the [test-latex-variable](https://github.com/showyourwork/test-latex-variable) repo:
image

The icon links to [here](https://github.com/showyourwork/test-latex-variable/blob/b92e0dc9cf0f2183e993b09f75dfe7ba31d4dd50/Snakefile#L4):
image

Issue #99, opened 2022-05-26
Showyourwork conda fails with `conda_forge` and `strict` channel priority


farr:
My default settings in my .condarc were based on a recommendation from `conda_forge`:

```
channels:
  - conda-forge
  - defaults
channel_priority: strict
```

with these settings, there are terrible package conflicts when `showyourwork` tries to set up its internal conda environments (i.e. the first time I run `showyourwork build` on a fresh install.  If I remove them from my `.condarc`, then setting up the internal conda environments proceeds fine.  I suspect that this is because `showyourwork` doesn't invoke `conda` with a specific channel setting (e.g. `-c defaults`) that overrides the user, and somehow the `conda-forge` packages are not suitable for the install.  

For completeness, here are the errors I saw:

```shell
> conda activate showyourwork # I created a custom environment for showyourwork to live in
> showyourwork build
Creating a new conda environment in ~/.showyourwork/env...
[...]

PR #200, opened 2022-09-23
Upgrade snakemake to 7.14.0

### LINKED ISSUE #199: Changing rule params in Snakefile doesn't trigger re-run of rule

rodluger:
Fixes #199. This PR will likely break a lot of things on the showyourwork side because of various hacks we've implemented, so we should thoroughly test it before merging.

### DIFF
--- a/docs/faqs.rst
+++ b/docs/faqs.rst
@@ -461,3 +461,23 @@ local installation of showyourwork*.)
 
 So, if you run into this error, we recommend you download all required files directly from
 the journal and include them (making sure to ``git add`` them) in your `src/tex` folder.
+
+
+Branch rename failed
+--------------------
+
+In versions of ``showyourwork`` prior to ``0.3.2``, users may occasionally run into the
+following error when attempting to run a third party's workflow:
+
+.. code-block:: text
+
+  Fetching Overleaf repo...
+  error: refname refs/heads/master not found
+  fatal: Branch rename failed
+
+This is a bug in ``showyourwork`` related to the fact that the default git branch on Overleaf
+projects is called ``master``, while the default branch on GitHub is called ``main``. This
+isn't an issue unless users don't have the correct credentials to access an Overleaf repository,
+in which case the ``git clone`` silently fails and no ``master`` branch is created.
+If you run into this error, simply delete or comment out the ``overleaf:`` section of the ``environment.yml``
+workflow config and re-run the workflow.
--- a/docs/layout.rst
+++ b/docs/layout.rst
@@ -389,14 +389,26 @@ build your article. You can read more about environment files
 By default, only the bare minimum specs are included
 (e.g., ``numpy`` and ``matplotlib``). Feel free to manually add to this list
 (noting that packages that can only be installed via ``pip`` should be placed in
-the ``pip`` section). It's recommended to either pin a specific version
-(i.e., ``matplotlib==3.3.4``) or specify a minimum version
-(i.e., ``matplotlib>=3.0.0``) for your packages. Just be aware that overconstrained
-requirements may break on other platforms
+the ``pip`` section). For packages listed under ``dependencies``, it is highly
+recommended that users specify **explict channels** and pin **exact versions**, e.g.:
+
+.. code-block:: text
+
+  dependencies:
+    - conda-forge::python=3.9
+
+For ``pip`` dependencies, users should also pin exact versions, e.g.:
+
+.. code-block:: text
+
+  - pip:
+    - matplotlib==3.3.4
+
+Note that overconstrained requirements may break the installation on other platforms
 (see `this post <https://stackoverflow.com/questions/39280638/how-to-share-conda-environments-across-platforms>`_),
-so you should probably only pin the direct dependencies of your project.
-If you alread have a ``conda`` environment for your project, you can export
-these direct dependencies -- the ones that you explicitly installed in the enviornment --
+so users should consider only pinning the *direct* dependencies of their project.
+If a ``conda`` environment already exists for a project, one can export
+these direct dependencies -- the ones whose installation was explicitly requested in the enviornment --
 by running
 
 .. code-block:: bash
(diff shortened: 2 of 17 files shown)


rodluger:
The new version of `Snakemake` now explicitly warns the user that setting `channel_priority: false` for conda environments is not recommended for reproducibility reasons. We had to relax change this setting from `strict` to `false` because of #99, but I agree with `Snakemake` that this is not ideal.

I would like to go back to forcing `strict` priority. To that end, I've specified explicit channels for all packages in both the `showyourwork` environment file (`showyourwork/workflow/envs/environment.yml`) and the cookiecutter template repo (`showyourwork/cookiecutter-showyourwork/{{ cookiecutter-repo }}/environment.yml`). We should recommend that users do the same for all additional packages they need installed for their articles.

For users who want to pin _all_ dependencies by doing a `conda list`, they [should be able to provide the `--explicit` argument](https://docs.conda.io/projects/conda/en/4.13.x/user-guide/tasks/manage-environments.html#building-identical-conda-environments) to output the exact URL from which all packages were downloaded.

rodluger:
Now requiring a minimum version of `conda` (4.12.0), since we really haven't tested any of showyourwork on older versions.

rodluger:
OK, there's a pretty big issue with this approach. Recall that there are always two versions of `showyourwork` at play: the version installed locally (whatever the user installed using `pip`), and the version specified in the `showyourwork.yml` config file (whatever version the author of the workflow specified). These each get installed in their own conda environments: the former gets installed globally in `~/.showyourwork/env/` and is used only to spawn the workflow from the command line; the latter gets installed in `.snakemake/envs/` (relative to the repository root) and is the one doing all the heavy lifting (parsing the user config, adding showyourwork rules to the DAG, applying patches to Snakemake, etc).

However, there is only ONE version of Snakemake -- the one specified in the global environment (the first one I mentioned above). This PR upgrades that version of Snakemake to `v7.14.0`. However, note what I mentioned above -- the workflow-defined version of showyourwork is the one that _applies patches to Snakemake_, and prior to this PR, it's applying patches to (what it assumes is) Snakemake `v6.15.5`.

This leads to an issue in our patch to the `wait_for_files` Snakemake function [here](https://github.com/showyourwork/showyourwork/blob/96b4b07d7a1c43928e5a4d7f871150a49a1aec17/showyourwork/patches.py#L377). In old versions of Snakemake that function expects a keyword argument called `ignore_pipe`. In newer versions that argument has been renamed to `ignore_pipe_or_service`. Therefore, when we use a newer version of `showyourwork` (>= this PR) and try to run a workflow defining an older version (< PR), we get the following error during the DAG build step:

```
Setting up the workflow...
Building the workflow graph...
Traceback (most recent call last):
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/__init__.py", line 730, in snakemake
    success = workflow.execute(
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/workflow.py", line 1075, in execute
    raise e
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/workflow.py", line 1071, in execute
    success = self.scheduler.schedule()
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/scheduler.py", line 423, in schedule
    self._finish_jobs()
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/scheduler.py", line 526, in _finish_jobs
    self.get_executor(job).handle_job_success(job)
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 660, in handle_job_success
    super().handle_job_success(job)
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/executors/__init__.py", line 257, in handle_job_success
    job.postprocess(
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/jobs.py", line 1101, in postprocess
    self.dag.check_and_touch_output(
  File "/Users/rluger/.showyourwork/env/lib/python3.9/site-packages/snakemake/dag.py", line 565, in check_and_touch_output
    wait_for_files(
TypeError: wait_for_files() got an unexpected keyword argument 'ignore_pipe_or_service'
```

This is fundamentally a backwards-incompatibility issue, since there is hard-coded stuff in previous versions of `showyourwork` that clash with the version of Snakemake we're upgrading to.

@dfm Any ideas on how to fix this?

rodluger:
There is only a single place in `Snakemake` where the new keyword argument name is used:

https://github.com/snakemake/snakemake/blob/v7.14.0/snakemake/dag.py#L569

One approach is to patch the user-installed version of `Snakemake` prior to executing the workflow with something like a `try-except` block to try both keyword names. This is ugly and terrible, but I think it will work. The only catch is that we are currently invoking `Snakemake` from the command line, so we'll have to spawn a Python session, patch Snakemake, and then call its `main` function instead.

rodluger:
The monkeypatch-on-the-monkeypatch idea worked. Not proud of it though, since this is not a sustainable approach! 

Perhaps this PR is not worth it because of the need to apply this patch. Or perhaps we can force the locally installed version of showyourwork to match the version defined in the workflow. This may be a pain for users, since they'll have to potentially upgrade/downgrade showyourwork fairly often. But perhaps they could just do things in separate conda environments to get around that.

rodluger:
... And I'm now getting issues related to user-defined Snakefile rules. In previous versions of showyourwork we [explicitly add rule scripts as inputs](https://github.com/showyourwork/showyourwork/blob/96b4b07d7a1c43928e5a4d7f871150a49a1aec17/showyourwork/userrules.py#L59), but in the newer versions of Snakemake that is done automatically. The old monkeypatch therefore leads to an error when we try to add an input that has already been added.

I'm going to abandon this PR because this would require yet another monkeypatch. So it looks like we're stuck with Snakemake version `6.15.5` for the time being, and we should find a different way of addressing #199.

At some point in the future -- perhaps during the next showyourwork rewrite -- we should probably just fork Snakemake and properly do away with these monkeypatches. We'll have to explicitly make things not backwards compatible at that point: articles generated with showyourwork prior to a certain version won't be able to be generated with newer version. I think that's ok as long as we properly document it, give good error messages, and plan very carefully when and how we'll execute this.

PR #203, opened 2022-09-28
Fork snakemake

rodluger:
I'm not sure if this is going to work yet, but I want to try to fork Snakemake instead of applying all the current monkey patches during runtime on the showyourwork side. This is, of course, the better approach, but the issue is that whatever we do we need to make backwards-compatible with versions of showyourwork <= 0.3.1, which apply monkeypatches assuming the version of Snakemake that is installed is 6.15.5 (making it very, very difficult to upgrade Snakemake -- see #200). This was a poor design decision early on, but I'm afraid we're stuck with the consequences.

More on this soon.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -573,33 +573,6 @@ repository. Exact names are required; no glob syntax allowed.
             - src/tex/figures
 
 
-.. _config.require_inputs:
-
-``require_inputs``
-^^^^^^^^^^^^^^^^^^
-
-**Type:** ``bool``
-
-**Description:** If there is no valid rule to generate a given output file
-(because of, e.g., a missing input file), but the output file itself is present on disk,
-Snakemake will not by default raise an error. This can be useful for running
-workflows locally, but it can compromise the reproducibility of a workflow when
-a third party attempts to run it. Therefore, the default behavior in |showyourwork|
-is to require all output files to be programmatically generatable when running
-the workflow, *even if* the output files exist on disk already. Otherwise, an
-error is thrown. Set this option to ``false`` to override this behavior.
-
-**Required:** no
-
-**Default:** ``true``
-
-**Example:**
-
-.. code-block:: yaml
-
-    require_inputs: true
-
-
 .. _config.run_cache_rules_on_ci:
 
 ``run_cache_rules_on_ci``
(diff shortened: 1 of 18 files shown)


rodluger:
Note to self: `datrie` (a dependency of `snakemake`) does not have pre-compiled wheels for Python 3.9 on PyPI, so I'm adding it as a `conda-forge` dependency in the internal `showyourwork` `environment.yml` file.

rodluger:
Ok, this is currently working -- I forked Snakemake [here](https://github.com/showyourwork/snakemake) and currently pinned the version used in `showyourwork` to the main branch of my fork. All tests pass for workflows created using the same version (i.e., that defined in this PR) of `showyourwork`.

The issue now is to figure out how to make this backwards compatible with articles based on previous versions of `showyourwork`. More on this soon.

rodluger:
#205/#206 is going to help with our backwards compatibility tests. This PR will almost certainly cause failures on `0.3.1` once #205/#206 is merged, so everyone please hold your breath.

rodluger:
Ok, we're finally at the stage where we can reliably test this PR. As expected, the latest run of the integration tests succeeded for a workflow based on the same version, but failed for a workflow based on the earlier `0.3.1` version (see [here](https://github.com/showyourwork/showyourwork/actions/runs/3152510440)) because of the incompatible monkeypatches discussed [in this thread](https://github.com/showyourwork/showyourwork/pull/200#issuecomment-1259757992).

The plan from here on out: convert as many of the monkeypatches as we can in [showyourwork/patches.py](https://github.com/rodluger/showyourwork-dev/blob/d18958ff7e5a0f6b040a5790b8e1f1461c2081e8/showyourwork/patches.py) to modifications to our [Snakemake fork](https://github.com/showyourwork/snakemake) to make our lives easier in the future. This won't solve any current problem, but I think now is a good time to start doing things properly.

Then, we need to make a decision regarding how to support previous versions of the workflow. We can either try to monkeypatch the existing monkeypatches in `0.3.x` or we can modify our fork of Snakemake such that those original monkeypatches don't lead to errors. I prefer the latter approach -- we can basically just edit the fork to make the breaking changes backwards-compatible (e.g., by allowing calls to [`wait_for_files`](https://github.com/showyourwork/snakemake/blob/76e0e6904f1f94ff4908bb10c1b85b9fdf6ae598/snakemake/io.py#L837) to pass _either_ `ignore_pipe` _or_ `ignore_pipe_or_service` keywords.

rodluger:
Hoorah! Things are now backwards compatible -- all tests passing. Now, let's start getting rid of the monkeypatches in favor of changes to the code in our Snakemake fork.

rodluger:
Closing in favor of #212 #213.

PR #204, opened 2022-09-28
Allow users to disable cache

rodluger:
This PR allows users to set `cache: false` in the `showyourwork.yml` config file to disable the Zenodo/Zenodo Sandbox caching functionality. Useful when, e.g., Zenodo Sandbox is down (as is the case right now -- I'm experiencing super high latency, which reaaaaally slows down the build).

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -14,6 +14,27 @@ customize several aspects of the workflow. Below is a list of all
 available options.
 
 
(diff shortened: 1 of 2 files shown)


rodluger:
Renamed setting to `cache_on_zenodo` since this setting does not affect the _local_ cache.

PR #205, opened 2022-09-28
Matrix strategy for workflow versions

rodluger:
Introduces functionality that lets us check for backwards-compatibility. Integration tests are now run using the current version of showyourwork to compile articles requesting (in `showyourwork.yml`) either the current version of the code or any previous version (defined in a build `matrix` in the Actions workflow config).

### DIFF
--- a/.github/workflows/local_integration_tests.yml
+++ b/.github/workflows/local_integration_tests.yml
@@ -11,6 +11,11 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     name: Run local integration tests
+    strategy:
+      matrix:
+        workflow-version:
+          - ${{ github.sha }}
+          - 0.3.1
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -32,4 +37,4 @@ jobs:
 
       - name: Run local integration tests
         shell: bash -l {0}
-        run: python -m pytest tests/integration
+        run: python -m pytest tests/integration --workflow-version ${{ matrix.workflow-version }}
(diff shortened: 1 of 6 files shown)


rodluger:
This seems to be working as expected! In fact, [this build](https://github.com/showyourwork/showyourwork/actions/runs/3144998134/jobs/5111730419) failed against `0.3.1` when running `test_letter.py` -- which was a unit test introduced recently after we patched an issue with the `letter` `documentclass` in `<=0.3.1`. I added an `xfail` to suppress this error (which again, means the workflow is behaving properly!)

rodluger:
@dfm I changed the way forks get tested to make the `matrix` approach work. Users can now provide a fork and ref when setting up a repo from the command line:

```bash
showyourwork setup user/repo --version https://github.com/user/forked-showyourwork-repo@ref
```

so there's [no need to edit the yaml file after it's created](https://github.com/showyourwork/showyourwork/blob/c4c596bf6508861c84d1e663aaad08cab9d3a94f/tests/integration/helpers/temp_repo.py#L163). Instead, we simply pass the version to the tests via a new argument [here](https://github.com/showyourwork/showyourwork/pull/205/files#diff-82f7b2d77c9380c2589aeeb7f866a3da918464833ffe9c6efa9e791b9d88afbcR50). For forks, [this](https://github.com/showyourwork/showyourwork/pull/205/files#diff-82f7b2d77c9380c2589aeeb7f866a3da918464833ffe9c6efa9e791b9d88afbcR21) is how I select the current version in the matrix.

Currently the matrix tests the current version and `0.3.1`. We can update it in the future as we release new versions.

rodluger:
Merging this and starting a new PR from staging -> main (#206).

PR #206, opened 2022-09-28
Matrix strategy for workflow versions

rodluger:
Rebases #205 to the `main` branch.

### DIFF
--- a/.github/workflows/local_integration_tests.yml
+++ b/.github/workflows/local_integration_tests.yml
@@ -11,6 +11,11 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     name: Run local integration tests
+    strategy:
+      matrix:
+        workflow-version:
+          - ${{ github.sha }}
+          - 0.3.1
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -32,4 +37,4 @@ jobs:
 
       - name: Run local integration tests
         shell: bash -l {0}
-        run: python -m pytest tests/integration
+        run: python -m pytest tests/integration --workflow-version ${{ matrix.workflow-version }}
--- a/.github/workflows/local_integration_tests.yml
+++ b/.github/workflows/local_integration_tests.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Run local integration tests
     strategy:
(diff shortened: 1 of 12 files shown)


rodluger:
If the Actions workflow is triggered from a `push` event, the `pull_request` context is empty and the version specification fails [here](https://github.com/showyourwork/showyourwork/pull/206/files#diff-82f7b2d77c9380c2589aeeb7f866a3da918464833ffe9c6efa9e791b9d88afbcR21) (it thinks the version we want is "@"), leading to a [failed check](https://github.com/showyourwork/showyourwork/actions/runs/3145377813/jobs/5112584818).

rodluger:
The passing tests are misleading: remote tests for PRs into the `staging` branch fail on startup because of a syntax error: https://github.com/showyourwork/showyourwork/actions/runs/3146197026/workflow

Looking into this...

rodluger:
Closing this for the moment to stop running tests while I debug.

rodluger:
Ok, the syntax error is fixed. I also added a `max-parallel: 1` to the `strategy` to avoid race conditions when pushing to the remote repos during the remote integration tests. Overall things are working as expected, with just one catch: the showyourwork-action installs the [latest pip-installable version of showyourwork](https://github.com/showyourwork/showyourwork-action/blob/fd5098db9c84dfd8d6b7adc27b4974088f64f895/src/conda.js#L54) (currently 0.3.1), which does not accept the new `--version <fork>@<repo>` syntax (introduced in this PR) when setting up a new repo -- this currently leads to several test failures.

rodluger:
[showyourwork-action@v1](https://github.com/showyourwork/showyourwork-action/tree/v1) has now been updated to allow users to choose a specific version of `showyourwork` to install.

rodluger:
Ok, I think this is mostly ready to be re-opened. The new behavior is for tests to always request that `showyourwork-action` install the **current** version of showyourwork (defined by the current commit / PR / whatever). The default for users is still to install the latest version on PyPI -- this change only affects the test suite.

rodluger:
PR #210 was a test of the remote integration testing workflow, and [everything worked as expected](https://github.com/showyourwork/showyourwork/actions/runs/3151826775/jobs/5126275924).

PR #212, opened 2022-10-03
Consistent environments & Snakemake upgrade

### LINKED ISSUE #126: Cache multiple conda environments
``showyourwork`` uses a special ``conda`` environment located in ``~/.showyourwork`` to spawn all workflows for a given user. This environment contains only ``showyourwork``-defined packages, so as long as a user uses the same version of ``showyourwork`` for all their projects, this environment doesn't have to be updated. However, every time a project defining a different version of ``showyourwork`` (in the ``.yml`` config file) is built, the environment in ``~/.showyourwork`` gets updated. This could get very annoying for users maintaining projects that depend on two or more versions of ``showyourwork``, as this introduces a ton of overhead every time they build.

An alternative is to store multiple ``conda`` environments in ``~/.showyourwork``, say up to five. If a workflow run would lead to the creation of a sixth, we simply delete the oldest one.

rodluger:
This is an alternative solution to that provided in #203. The source of all our problems with backwards compatibility is the fact that we are using this [environment file](https://github.com/showyourwork/showyourwork/blob/22cbcdfca0c77c0fbd4e929f2f29dc2a52274eb2/showyourwork/workflow/envs/environment.yml) when setting up the conda environment for the article workflow. This is a problem because the article defines its own version of `showyourwork` (in the `showyourwork.yml` file), but we're using the environment specs defined in the _currently installed_ version of `showyourwork`. When these differ, we can run into issues like the ones discussed in #203.

Rather than fork Snakemake and monkeypatch our way to backwards compatibility, I suggest we instead focus on installing the correct dependencies in the first place. I guess this sound obvious in hindsight, but it's going to require a little sleight of hand to implement.

More soon.

### DIFF
--- a/docs/faqs.rst
+++ b/docs/faqs.rst
@@ -122,9 +122,9 @@ incomplete. When this happens, ``showyourwork`` tells the user:
 
   If you are sure that certain files are not incomplete, mark them as complete with
 
-    showyourwork build --cleanup-metadata <filenames>
+    showyourwork --cleanup-metadata <filenames>
 
-  To re-generate the files rerun showyourwork build with the --rerun-incomplete flag.
+  To re-generate the files rerun your command with the --rerun-incomplete flag.
 
 Sometimes, however, the ``--cleanup-metadata`` argument does not successfully
 clean up the incomplete files. This may be due to either an issue with Snakemake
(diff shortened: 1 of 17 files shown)


rodluger:
I think the cleanest way of doing this is to make `showyourwork` available on `conda-forge` so we can explicitly declare things like `snakemake` and `tectonic` as dependencies. Looking into that now.

rodluger:
Note to self: should we allow showyourwork to run a workflow defining a newer version of the code?

rodluger:
Remote tests failing because (as I keep forgetting), we run PRs in the context of the **base**, not the fork, and this PR changes some fundamental structural things about the code, leading to inconsistencies. All is well -- going to merge this into the `staging` branch so we can properly test it on `push` events.

PR continued in #213.

PR #213, opened 2022-10-05
Consistent environments & Snakemake upgrade

rodluger:
Rebases #212 onto the main branch.

In brief, this PR ensures that the version of `Snakemake` (and all other `showyourwork` dependencies) is determined by the version of `showyourwork` defined in the _workflow_, and not the currently installed version. This fixes several issues with backwards compatibility and allows us to, e.g., upgrade `Snakemake` without breaking everything.

### DIFF
--- a/docs/faqs.rst
+++ b/docs/faqs.rst
@@ -122,9 +122,9 @@ incomplete. When this happens, ``showyourwork`` tells the user:
 
   If you are sure that certain files are not incomplete, mark them as complete with
 
-    showyourwork build --cleanup-metadata <filenames>
+    showyourwork --cleanup-metadata <filenames>
 
-  To re-generate the files rerun showyourwork build with the --rerun-incomplete flag.
+  To re-generate the files rerun your command with the --rerun-incomplete flag.
 
 Sometimes, however, the ``--cleanup-metadata`` argument does not successfully
 clean up the incomplete files. This may be due to either an issue with Snakemake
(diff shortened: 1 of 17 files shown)


rodluger:
Although all tests are currently passing (unit, local, and remote, both against current version and `0.3.1`), there are still some wrinkles to work out.

1. We've tested backwards compatibility, but what about forwards compatibility? For example, if the user has `0.3.1` installed locally, what happens if they try to execute a workflow requesting `0.3.2`? It should be fine, right? _Right?_

2. The current approach to ensuring the correct dependencies are installed is hacky. When spawning the workflow, we [figure out the version](https://github.com/showyourwork/showyourwork/pull/213/files#diff-838c05bbb5f9e31298fb3a381b3a1e5f0a3fdc7231d1db9793e99601ba4a67a3R97) of `showyourwork` to use from the repo's config file, and then [download the `environment.yml` and `.condarc` files from the corresponding ref](https://github.com/showyourwork/showyourwork/pull/213/files#diff-c22c56e01ddc44bcd491ac389b80fe3d44a267bb5b75f8b5f7e42aee417f1362R126-R146) on GitHub. We then create a new `conda` environment from those specs and spawn the workflow within it. It's not enough to just `pip install` the correct version of `showyourwork` within the new environment, since we rely on several `conda-forge` dependencies (`tectonic`, `mamba`, etc). We might be able to simply put `showyourwork` on `conda-forge` to do away with this hack, but this won't work for workflows requesting versions of `showyourwork` defined on forks, installed locally, or provided as a specific git commit SHA. We might also run into trouble with backwards compatibility, again.

I need to keep thinking about all this. My feeling is that we're quickly approaching a point where all these structural changes justify a (partial?) rewrite of the code and the release of `0.4.0` (or, if we're feeling adventurous, `1.0.0`), at which point we can explicitly forgo backwards compatibility.

rodluger:
@dfm @adrn I'd love to chat about all this, and the future of the code, when you both have time.

rodluger:
Moved to https://github.com/rodluger/showyourwork-dev/pull/2 as I consider a different approach.

Issue #216, opened 2022-10-08
Tectonic error


rodluger:
Workflows may fail with

```text
Setting up the workflow...
error: error sending request for url (https://relay.fullyjustified.net/default_bundle.tar): error trying to connect: dns error: failed to lookup address information: Name or service not known
caused by: error trying to connect: dns error: failed to lookup address information: Name or service not known
caused by: dns error: failed to lookup address information: Name or service not known
caused by: failed to lookup address information: Name or service not known

An error occurred while compiling the manuscript.
```

This is an issue with one of the servers hosting the `tectonic` bundle; see [here](https://github.com/tectonic-typesetting/tectonic/issues/949). Hopefully this is temporary -- if not, I'll look into a workaround.

rodluger:
For those in need of a temporary workaround, try adding the following line to the `showyourwork.yml` config file:
```yaml
tectonic_args: ["-w", "https://archive.org/services/purl/net/pkgwpub/tectonic-default"]
```
which tells `tectonic` to use a different server for the bundle.

rodluger:
Issue was due to an expired registration; now fixed (https://github.com/tectonic-typesetting/tectonic/issues/949).

Issue #215, opened 2022-10-07
AttributeError: 'str' object has no attribute 'name'


rodluger:
Many workflows may start failing soon with the following [error](https://github.com/showyourwork/test-default/actions/runs/3206142982/jobs/5239520715):

```
Traceback (most recent call last):
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/snakemake/__init__.py", line 701, in snakemake
    success = workflow.execute(
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/snakemake/workflow.py", line 1060, in execute
    logger.run_info("/n".join(dag.stats()))
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/snakemake/dag.py", line 2191, in stats
    yield tabulate(rows, headers="keys")
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/tabulate/__init__.py", line 2048, in tabulate
    list_of_lists, headers = _normalize_tabular_data(
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/tabulate/__init__.py", line 1471, in _normalize_tabular_data
    rows = list(map(lambda r: r if _is_separating_line(r) else list(r), rows))
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/tabulate/__init__.py", line 1471, in <lambda>
    rows = list(map(lambda r: r if _is_separating_line(r) else list(r), rows))
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/tabulate/__init__.py", line 107, in _is_separating_line
    (len(row) >= 1 and row[0] == SEPARATING_LINE)
  File "/home/runner/.showyourwork/conda/d9b5401cad5d7920c0bc8d4a79457fac/lib/python3.9/site-packages/snakemake/rules.py", line 1138, in __eq__
    return self.name == other.name and self.output == other.output
AttributeError: 'str' object has no attribute 'name'
```

[...]

PR #217, opened 2022-10-08
Fix envs

### LINKED ISSUE #199: Changing rule params in Snakefile doesn't trigger re-run of rule

### LINKED ISSUE #214: How do we maintain older versions of showyourwork?
An update today to the `tabulate` package (`v0.8.10` -> `v0.9.0`) breaks `Snakemake` `v7.14.2` because of [this line](https://github.com/snakemake/snakemake/blob/v7.14.2/snakemake/rules.py#L1222) in the `__eq__` method of a `Rule`. 

https://github.com/snakemake/snakemake/issues/1892

Basically when building up the summary of the DAG, at some point `tabulate` checks whether an instance of a rule is equal to a certain string, which that `__eq__` is not capable of handling, causing an `AttributeError` instead of just returning `False`. This happens at the beginning of the workflow so it causes _everything_ to fail, for all builds of all `showyourwork` projects. I suspect it might also break previous versions of `Snakemake`, but I need to test that still. 

It's a trivial fix on the `Snakemake` side -- e.g., just check if the RHS and the LHS are both of type `Rule`. On our end, we can fix things for the next release of `showyourwork` by monkeypatching `Snakemake` or simply pinning `tabulate` at `v0.8.10` (that's what I had to do in #213, which upgrades the workflow to use this newer version of `Snakemake`).

But how do we fix already released versions of `showyourwork`? Workflows based on `0.3.1` and earlier will begin to fail with a cryptic `AttributeError` deep inside `tabulate`. The usual approach to this is to tell users to upgrade -- I imagine that's what `Snakemake` will have to do.

But our problem is more serious: our workflows are defined to work with _specific_ versions of `showyourwork` (and, by extension, `Snakemake`), since we want things to be reproducible. But because `Snakemake` doesn't pin all of its dependencies at exact versions (and those dependencies don't pin _their_ dependencies, etc.), we're running into a **fundamental** problem with this approach: things can break at any time, without warning, and there's nothing we can do about it other than tell users to manually bump the version of `showyourwork` defined in a workflow.

In my mind, the only solution is to reconsider the versioning of `showyourwork` workflows. Perhaps we should always build articles with the latest version and abandon the `version:` syntax in the config file? We'd have to be very careful about backwards compatibility (as always).

@dfm what do you think?

rodluger:
This PR will revamp our approach to handling `conda` environments and dependencies when building `showyourwork` articles. Supercedes #200, #203, #212, #213.

Currently (`v0.3.x`), running the locally installed `showyourwork` ("local version") in an article repository creates a new `conda` environment and in it installs the version of `showyourwork` defined in the repo's config file ("target version"). The workflow is then executed in that isolated environment. This approach needs to be rethought for a variety of reasons, including:

1. We made a pretty bad decision early on concerning how the `conda` environment is configured: we create it based on an `environment.yml` file in the _local version_ of `showyourwork`, not in the _target version_ of `showyourwork`. This led to inconsistent environments. The biggest problem is that `showyourwork` applies some [monkeypatches](https://github.com/showyourwork/showyourwork/blob/v0.3.1/showyourwork/patches.py) to `Snakemake` and some of the functions we're patching changed in newer versions of `Snakemake`. This makes it virtually impossible for us to upgrade the pinned version of `Snakemake` (`6.15.5`) without breaking someone's build. For example, if a user runs `showyourwork>0.3.1` to build an article requesting `showyourwork<=0.3.1`, the workflow will apply the _old_ monkeypatches, which are not compatible with newer versions of `Snakemake`, leading to all sorts of errors.

2. There's an even more fundamental issue with our approach that makes `showyourwork` articles _all but guaranteed to start failing_ one day or another. Currently (`v0.3.x`), article workflows define exact versions of `showyourwork` to be run with. The idea here was to ensure reproducibility -- while the code may change, leading to differences in the output of workflows, if we required articles to pin `showyourwork` we could guarantee the output remained constant. This was the idea behind installing the "target version" of `showyourwork` in an isolated environment -- so we could _exactly_ reproduce the environment in which the paper was originally built. However, this approach is seriously flawed because we chose not to pin _all_ dependencies, only the few _direct_ dependencies of `showyourwork` (see the [environment.yml](https://github.com/showyourwork/showyourwork/blob/v0.3.1/showyourwork/workflow/envs/environment.yml) file we use to create the `conda` environment). We decided this because it is currently [extremely difficult (impossible?)](https://stackoverflow.com/questions/39280638/how-to-share-conda-environments-across-platforms) to specify complete environments in a platform-independent way with `conda`. However, our approach came back to haunt us this week, when an innocuous update to the `tabulate` package led to errors in `Snakemake` (which does not pin tabulate -- so all new installs of pretty much any version of `Snakemake` are now broken; see #215). This has since been [patched](https://github.com/snakemake/snakemake/releases/tag/v7.15.2) in `Snakemake==7.15.2`, so if `showyourwork` were a normal package we could just upgrade our pinned version, tell our users to update, and move on. But we can't do that, as per point 1 above!

We seriously need a new approach -- and soon. I released postfixes (`0.3.x.post1`) to address the `tabulate` issue, but I think we've reached the point where we need to rethink the structure of the code.

As I see it, we have two options:

A. Switch over to fully constrained environment specifications. I don't think there's a way to guarantee compatibility across all platforms, but perhaps there are ways to generate multiple `environment.yml` files for all the main platforms out there. It still doesn't seem future-proof to me -- e.g., would the release of the Apple M1 chips screwed everything up if we'd done this a couple years ago? Also, ironically (but perhaps fortuitously), `tectonic` is no longer working as of this morning (#216) because the server hosting its bundles is down. Hopefully that will resolve itself, but what if the developers were to decide to switch to a new server and stop supporting the current one? Then all previous versions of `tectonic` would stop working, regardless of how good a job we did at ensuring a consistent environment. I don't necessarily expect this to happen with `tectonic`, but in general I think that this whole approach places too much trust in the developers of the hundreds (thousands?) of indirect dependencies of `showyourwork`. Who knows how many of those packages depend on external services that could shut down at any moment? I don't want to corner us into more problems down the line because of this.

B. Abandon the idea that articles need to be compiled with specific versions of `showyourwork`. The idea there was to enforce reproducibility, but we've now seen that this comes with its own can of worms. I suggest we take the `Snakemake` approach: Snakefiles don't define a specific version they need to be run with -- only a _minimum version_. The onus then is on `Snakemake` to ensure backwards compatibility of their code. I think we can work with that model, at least going forward. The idea then would be to do away with the multiple versions of `showyourwork` (the "local" and "target" versions described above) and always build workflows with the currently installed version. We'd then take the standard software package approach of telling users to upgrade if things are broken. This makes it MUCH easier to release patches -- solving points 1 and 2 above is much, much easier.

This PR will implement option B. It's still a work in progress, so please stay tuned -- and feel free to give me feedback!


rodluger:
Paging @dfm for input when you have some time.

dfm:
I haven't looked in detail, but here are some initial impressions:

> A. Switch over to fully constrained environment specifications. ...

For this approach, the [conda-lock](https://github.com/conda-incubator/conda-lock) project is designed to do exactly this: create a set of platform-specific lock files.

> B. Abandon the idea that articles need to be compiled with specific versions of showyourwork. ...

I think this is probably a good approach for `showyourwork`'s environment, but I think there's probably something to be said for locking the environment used to generate the figures, etc., since we don't have control over making sure that (for example) numpy and its dependencies remain backwards compatible with this same level of strictness that you're proposing here.

So, my initial gut feeling is that we should implement option B for the process that drives show your work, and then perhaps use `conda-lock` for the actual figures. That being said, that's probably something that should be part of Snakemake and the fact that it doesn't currently exist probably has a good reason. I'll keep thinking and lets chat some more.

dfm:
Oh - Snakemake actually does have support for this via conda "pins" https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#freezing-environments-to-exactly-pinned-packages

We could probably generate the pins using `conda-lock`.

rodluger:
Continued in #221.

PR #221, opened 2022-10-10
Better approach to showyourwork envs and deps

### LINKED ISSUE #126: Cache multiple conda environments

### LINKED ISSUE #186: Add all config options to the cookiecutter template
This was suggested by @MilesCranmer. This entails adding all of the [config options](https://github.com/showyourwork/showyourwork/blob/cbbd33da0925371441e9a4a94b1b0bcca105ec5b/docs/config.rst) with their default values to [this file](https://github.com/showyourwork/showyourwork/blob/cbbd33da0925371441e9a4a94b1b0bcca105ec5b/showyourwork/cookiecutter-showyourwork/%7B%7B%20cookiecutter.repo%20%7D%7D/showyourwork.yml). An easy PR for anyone wanting to get involved.

### LINKED ISSUE #199: Changing rule params in Snakefile doesn't trigger re-run of rule

### LINKED ISSUE #214: How do we maintain older versions of showyourwork?

### LINKED ISSUE #215: AttributeError: 'str' object has no attribute 'name'

rodluger:
Rebases #217 onto main.

Fixes https://github.com/showyourwork/showyourwork/issues/126. Fixes https://github.com/showyourwork/showyourwork/issues/199. Closes https://github.com/showyourwork/showyourwork/issues/214. Fixes https://github.com/showyourwork/showyourwork/issues/215. Closes https://github.com/showyourwork/showyourwork/issues/186.


rodluger:
The [failure](https://github.com/showyourwork/showyourwork/actions/runs/3227675733/jobs/5282791861) is expected: remote tests [fail with the error `snakemake: not found`](https://github.com/showyourwork/test-matplotlib-latex/actions/runs/3227795565). That's because the `showyourwork-action` is still trying to install `showyourwork` with `pip`, so `snakemake` doesn't get installed. I'm going to publish a pre-release on `conda-forge` and switch to a `conda install` in `showyourwork-action`.

rodluger:
I realized we actually don't need to put `showyourwork` on `conda-forge` at all. The only non-python deps we truly need are `tectonic` and `graphviz`, and these are needed only within certain Snakemake rules in the workflow. We can therefore let `Snakemake` install them for us. It used to be the case (in the previous version of Snakemake we'd pinned) that `mamba` was needed to run `Snakemake` (meaning we'd have to list `mamba` as a dependency of `showyourwork`), but this has since been fixed -- if we pass `--conda-frontend conda` to `snakemake` we can get away without the need to install `mamba`, and we can just install `snakemake` itself (technically `snakemake-minimal`) from `pip`.

PR #230, opened 2022-10-25
copy synctex file over

### LINKED ISSUE #229: Using SyncTeX
Hey @rodluger et al,

I was wondering if it's possible to use SyncTeX at all? It seems like there isn't a synctex file generated during the build process. 

In particular I'm using the vscode version - not sure if there's any extra config required given that ms.pdf and ms.tex are in different folders.

Thanks!
Miles

rodluger:
Fixes #229.

### DIFF
--- a/showyourwork/workflow/scripts/pdf.py
+++ b/showyourwork/workflow/scripts/pdf.py
@@ -90,4 +90,4 @@ if __name__ == "__main__":
         path = paths.user().compile / (
             snakemake.config["ms_name"] + ".synctex.gz"
         )
-        shutil.copy(str(path), ".synctex.gz")
+        shutil.copy(str(path), snakemake.config["ms_name"] + ".synctex.gz")
--- a/showyourwork/workflow/scripts/pdf.py
+++ b/showyourwork/workflow/scripts/pdf.py
@@ -90,4 +90,4 @@ if __name__ == "__main__":
         path = paths.user().compile / (
             snakemake.config["ms_name"] + ".synctex.gz"
         )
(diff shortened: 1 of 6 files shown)


showyourwork/workflow/scripts/pdf.py
```diff
@@ -78,3 +78,9 @@
         str(paths.user().compile / (snakemake.config["ms_name"] + ".pdf")),
         str(Path(snakemake.config["ms_pdf"])),
     )
+
+    # Copy the synctex file (if present) to the user dir
+    for file in [".synctex", ".synctex.gz"]:
```
MilesCranmer: I would check the .gz file first, and, if it succeeds, break. The reason is that there may be both files - the .gz file should be preferred since it’s smaller.

MilesCranmer (commented):
Looks good. My one other comment is that you might want to make this default, since it shouldn’t be very expensive but is quite useful.

showyourwork/workflow/scripts/pdf.py
rodluger: Good idea!

rodluger:
OK, these changes make `--synctex` the default option. Now we're also marking the `synctex` file an explicit output of the build rule, so running `showyourwork clean` will delete it.

rodluger:
Converting this PR to a draft while I debug the errors.

rodluger:
Ready for review!

MilesCranmer (approved):
Looks good to me! Thanks for doing this.

PR #244, opened 2022-12-13
Fix `Illegal parameter` error in minimal `showyourwork.sty`

MilesCranmer:
Right now, compiling the Overleaf equivalent of a showyourwork repo, you will see the (non-breaking) error:
```
Illegal parameter number in definition of \reserved@a.
<to be read again> 
```
The solution is to use `##1` instead of `#1` in command definitions.

See https://tex.stackexchange.com/a/639980: 

> \IfFileExists doesn't work like “usual” conditionals: it stores the true and false branches in macros, and uses them later. This requires you to double every # in there:

### DIFF
--- a/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/src/tex/showyourwork.sty
+++ b/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/src/tex/showyourwork.sty
@@ -9,5 +9,5 @@
     \newcommand\GitHubIcon{}
     \newcommand\showyourwork{}
     \newcommand\script[1]{}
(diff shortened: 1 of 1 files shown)


MilesCranmer:
@dfm want to give this a review+merge?

dfm:
Sorry about the delay! Just catching up on GitHub notifications after a crazy December.

This looks good to me. Besides the emitted error notification, did this cause other issues in the build?

I'm running the full test suite now - happy to merge as soon as those pass.

MilesCranmer:
I don’t remember but I think Overleaf’s latexmk settings were able to ignore it and compile normally, whereas my local build simply refused to do so.

Issue #246, opened 2022-12-16
Docs suggestion: How to use rule outputs for directories cached on Zenodo


LouisDesdoigts:
In the [Zenodo section](https://show-your.work/en/latest/zenodo/) of the docs it shows how to cache a directory to Zenodo, but doesn't show how to update the `Snakefile` and `showyourwork.yml` to access these directories appropriately. I ended up looking through the projects that use showyourwork to find out how to do this, ie

In Snakefile
```
rule:
    input:
        rules.some_other_rule.output
```

and in the `showyourwork.yml`
```
dependencies:
  src/scripts/some_script.py:
    - src/data/some_directory/
```

I think this could be quite valuable for other users like me who are new to snakemake etc. 

I would be happy to make this update to the docs myself in a branch and submit it for review, so let me know you think that would be helpful!

Cheers

dfm:
Is that not what is described here: https://show-your.work/en/latest/zenodo/#the-streamlined-version

What is missing from that discussion and where do you think it should go? I'd be very happy to review a PR that makes this clearer!

dfm:
Closed by #248, I think!

PR #248, opened 2023-01-04
Updating the zenodo docs to detail how to access cached directories

LouisDesdoigts:
Just a small update to the docs to be more explicit about how to update the rules and dependencies when working with cached directories in Zenodo, as discussed in #246.

### DIFF
--- a/docs/zenodo.rst
+++ b/docs/zenodo.rst
@@ -333,7 +333,7 @@ users with access to your account can see their files.
                 "src/figures/figure.pdf"
             script:
                 "src/scripts/figure.py"
-    
+
     You will also need to update the ``showyourwork.yml`` config file to reflect
     the new dependency.
 
--- a/docs/zenodo.rst
+++ b/docs/zenodo.rst
@@ -321,7 +321,7 @@ users with access to your account can see their files.
     directory and upload the archive to Zenodo Sandbox, then download and unzip
     it the next time it is needed.
 
-    Scripts that use the files in this output directory can have their rule 
+    Scripts that use the files in this output directory can have their rule
     inputs point directly to the directory itself.
 
     .. code-block:: python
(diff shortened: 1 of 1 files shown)


docs/zenodo.rst
```diff
@@ -321,6 +321,30 @@ users with access to your account can see their files.
     directory and upload the archive to Zenodo Sandbox, then download and unzip
     it the next time it is needed.
 
+    You will also need to update the rule for any scripts that use the output
```
dfm: This is only necessary for the cases where you've already written a custom Snakemake rule for your script, right? The default rules will handle this properly as long as you update `showyourwork.yml`, right?

If so, it might be better to make that more explicit in this comment since I think it's a bit confusing right now…

docs/zenodo.rst
```diff
@@ -321,6 +321,30 @@ users with access to your account can see their files.
...
+        rule figure:
+            input:
+                rules.simulation.output
+            output:
+                "src/figures/figure.pdf"
+            script:
+                "src/scripts/figure.py"
+
+    You will also need to update the ``showyourwork.yml`` config file to reflect
+    the new dependency.
```
dfm: I'd probably remove this next block and just say something like "don't forget to update your `showyourwork.yml`", since this syntax is already discussed above.

dfm:
Thanks @LouisDesdoigts!! I asked some tiny questions above.

docs/zenodo.rst
LouisDesdoigts: Ah yes I didn't think of the case where the inputs are not generated by your own rules. 

I did some more testing and found the syntax to be much more flexible than I had initially realised (I must have had other bugs during my initial tests). I will update this section to remove the rule output and show that you can point to the directory.

docs/zenodo.rst
LouisDesdoigts: I've found that you need to explicitly point to the *directory* rather than individual files if using the `directory(...)` output. This section only shows dependencies on individual files, which won't work from what I have found, so I think this might actually be valuable to detail explicitly? It might be obvious, but definitely caused me some trouble as someone who was coming with zero experience of these different systems.

dfm:
Thanks @LouisDesdoigts!

PR #251, opened 2023-01-10
Protected the \nolinenumbers command

### LINKED ISSUE #252: AASTEX631 included has a bug with lineno
The version of aastex631.cls provided by the cookie cutter template is different to the version actually used by tectonic. In the included version a \nolinenumbers command is not protected with the \caption definition and causes an error if you run tectonic within the 'tex' directory and don't use the lineno package. The aastex631.cls pulled by tectonic seems to be an older version (pre-October 21) with a different command at the start of the \caption definition.

https://github.com/showyourwork/showyourwork/blob/cce14c8a3a028d51640dfdb72047a0991435b66f/showyourwork/cookiecutter-showyourwork/%7B%7B%20cookiecutter.repo%20%7D%7D/src/tex/aastex631.cls#L6236

(compared to this from pre-Oct 21
```
\def\caption{\numlinesfalse                                                              
```
that seems to rely on a variable `linenumbers` from the `lineno` package.)

Fixed in https://github.com/showyourwork/showyourwork/pull/251 ,but this isn't part of the standard workflow and is only triggered if lineno is not used and tectonic is used within the tex directory. It might cause a problem on Overleaf or plain latex/xelatex, but not tested.

eelsirhc:
The \nolinenumbers command is defined by the lineno package that isn't required, added the conditional check for the \numlines flag defined when lineno is imported

### DIFF
--- a/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/src/tex/aastex631.cls
+++ b/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/src/tex/aastex631.cls
@@ -6233,7 +6233,7 @@ on page
 %% text to listoftables or listoffigures, which we are not going to use
 %% anyway. This change enables track changes commands to work in captions.
 \def\xtable{table}
-\def\caption{\ifnumlines\nolinenumbers\fi 
+\def\caption{\ifnumlines\nolinenumbers\fi
 \ifx\@captype\@undefined
 \@latex@error {\noexpand \caption outside float}\@ehd
 \expandafter \@gobble \else
--- a/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/src/tex/aastex631.cls
+++ b/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/src/tex/aastex631.cls
@@ -6233,7 +6233,7 @@ on page
 %% text to listoftables or listoffigures, which we are not going to use
 %% anyway. This change enables track changes commands to work in captions.
 \def\xtable{table}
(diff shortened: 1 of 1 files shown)


dfm:
@rodluger introduced this change to [handle some incompatibility between the margin icons and line numbers](https://github.com/showyourwork/showyourwork/commit/fe6c56d20304264ac5bdbf613685c821adc95890). I'm not sure what the source of that issue was or whether or not it's something we still need to worry about. Either way, unless @rodluger comments otherwise, I'd be happy to merge this because I'd rather make sure that SYW works with the out-of-the-box classfile.

eelsirhc:
It didn't occur to me that the change was specific to this package. I assumed it was a conflict between versions of TeX packages I have.

I *think* my change is equivalent to the [original patch](https://github.com/showyourwork/showyourwork/commit/fe6c56d20304264ac5bdbf613685c821adc95890). With the original `\numlinesfalse` (or no extra check) the caption command seems to put the GitHub logo in the wrong place when line numbers are used. The `\nolinenumbers` version always works within `showyourwork` because there's a hidden import of the lineno package (build.tex -> mathtools -> amsmath -> lineno) whether you use it or not, but breaks outside of the build environment without an import of mathtools/amsmath/lineno.

With my modification, it builds inside and outside of showyourwork, with and without the import. I don't know if my fix is the best of the options (a:don't run outside showyourwork, b:require an import with this aastex631 style,  or c: changing the aastex631 file) but I just copied the syntax from other parts of the aastex file so I think it's robust.

dfm:
Thanks @eelsirhc! Hearing no concerns, I'm going to go ahead and merge this.

PR #255, opened 2023-01-18
Update project using showyourwork

arm61:
Additionally, I realised I could put anything in the field, not just arXiv fields. 

Would there be interest in extending the `projects.rst.jinja` to support traditional reference too? I am happy to open a PR for it if so.

### DIFF
--- a/docs/projects.json
+++ b/docs/projects.json
@@ -95,7 +95,7 @@
       "Andrew J. Caruana",
       "Christy J. Kinane",
       "Alexander J. Armstrong",
-      "Tom Arnold",
+      "Thomas Arnold",
       "Joshaniel F. K. Cooper",
       "David L. Cortie",
       "Arwel V. Hughes",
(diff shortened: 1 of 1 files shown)


docs/projects.json
```diff
@@ -104,8 +104,11 @@
       "Wojciech Potrzebowski",
       "Vladimir Straostin"
     ],
-    "doi": "arXiv:2207.10406",
-    "url": "https://arxiv.org/abs/2207.10406",
-    "field": "Statistics"
+    "doi": "10.1107/S1600576722011426",
+    "url": "https://doi.org/10.1107/S1600576722011426",
+    "version": "1.0.0".
```
dfm: Looks like a syntax error here! `.` -> `,`

arm61: Ooops!

```suggestion
    "version": "1.0.0",
```

PR #256, opened 2023-01-18
add example paragraph for reproducibility

mathren:
as discussed at lunch, taken from my last accepted paper and provided as minimum example that can be edited by users.

Not sure this is the best place where to add this, though.

Also adds the paper from which this is taken to the list of projects

### DIFF
--- a/docs/latex.rst
+++ b/docs/latex.rst
@@ -261,9 +261,13 @@ A macro that resolves to the current commit SHA
 Reproducibility paragraph
 -------------------------
 
-Many scientific journals (e.g., those of AAS) will require the authors
-to add a paragraph or appendix describing in a few words how
-|showyourwork| operates and how the article can be reproduced.
+To explain to the readers of your article how |showyourwork| operates,
+how they can reproduce your science, and where to find the data and
+scripts used, it can be useful to have a brief paragraph or appendix
+in your manuscript. This is also often helpful to meet the
+reproducibility requirement of scientific journals and make your
+article self-contained. Typically this will also contain the doi of
+the zenodo repository associated to your article.
 
 The following is an example from the appendix E of `Renzo et al. 2023
 <https://ui.adsabs.harvard.edu/abs/2023ApJ...942L..32R/abstract>`_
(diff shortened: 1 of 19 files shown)


dfm:
cc @augustfly, @gregschwarz

augustfly:
I have a couple of comments that I'll put right into the commit (that's right right?)

augustfly:
And @mathren even if you have already done so, I'm sending an email to the ApJL production editor about their full time fuckup of the links in your article published last week. This is what we are trying to avoid doing, and is weird bc we were just discussing all this with our publishing partners so I have no idea what happened (and get it fixed straightaway).  All I can do is apologize. @gregschwarz

augustfly:
As a general comment I think it is worth asking if there is a location in your article that show-your-work users *want/prefer* this paragraph to appear. In @mathren's Letter is ended up in an appendix. Part of me is of the opinion it is more useful appearing right up front -- maybe right after the introduction. I'm tagging ppl from AAS241 who aren't yet on this thread but may have opinions (my tags being is an unintentional subset of anyone else who might have opinions) @afeinstein20 @jfcrenshaw

mathren:
@augustfly I actually had not noticed! I will also email them, and it's probably also on me not having checked thoroughly. Will email them as well if that is helpful to hammer the point.

Regarding the location in the paper: in my case it was a letter and so appendix was the obvious choice (since I was already far beyond the word-count). I also think this is info that is relevant to the subset of people that may try and reproduce the study, which hopefully is a (small) subset of the readers, so this is another argument in favor of putting this in an appendix. On the other hand, a short paragraph like this could also fit in the methods or end of introduction. If you converge on a specific recommendation we can definitely also mention it in the documentation!

gregschwarz:
We get SYW manuscripts quite regularly. It would be good to broaden the discussion to some of them so they can participate @bruce-edelman @bfarr

docs/latex.rst
```diff
@@ -256,3 +256,37 @@ A macro that resolves to the current repository URL
 
 A macro that resolves to the current commit SHA
 (i.e., ``31860f2f558b05d8c941d8f73c64f5dbf5ee79db``).
+
+
+Reproducibility paragraph
+-------------------------
+
+Many scientific journals (e.g., those of AAS) will require the authors
```
dfm: I'd say that it's less about the journals "requiring" it (although that may be a factor) and more about the fact that we (as the authors) want to share info about the mode we're working in. So I might re-phrase this intro...

dfm (commented):
I've added a minor comment about the wording, but I'm otherwise happy to merge this! I think in the long run we might want to think about where this should live and if we might want to have more specific instructions about where to put this as others have been discussing (another option might be a canned paragraph for the acknowledgements), but I'm inclined to merge this as a good solution in the short term!

augustfly:
The only change I really hope is made before this is committed is the Zenodo URL => DOI form!

dfm:
Thank you, @mathren!!

PR #266, opened 2023-01-27
Splitting monolithic rules to enable finer-grain control of tasks

### LINKED ISSUE #261: Switch to using built in Snakemake support for conda environments
We currently hackily construct our own conda activate command [here](https://github.com/showyourwork/showyourwork/blob/02a746eb844ee4fb054cc2cb1e4b1ef37c1bcff4/showyourwork/workflow/scripts/conda.py), which is then used in subsequent `subprocess` calls to activate the environment. This is necessary because currently some of the snakemake rules are too monolithic and actually require several different environments. We should be able to refactor these rules to split these subtasks out into independent jobs.

dfm:
Closes #261

I'm starting to work on refactoring the rules to get rid of the hacky "internal conda environment". ~~So far I have worked through `compile` as a proof of concept. Just need to do it for the preprocessing step now too.~~ Everything works now.

~~I'm planning to also add support for using `mamba` if it is installed, which I think isn't currently supported.~~ Edit: I did this in #268 instead.

One fairly large change with this is that we're no longer building the manuscript in place. Instead, we copy the entire contents of the `tex` directory (as well as various other style files, etc.) over to `.showyourwork/compile` and then build there. This made it easier to avoid drowning the working directory in temporary files, and I don't foresee any major issues, but maybe there's something I missed.


dfm:
@rodluger might or might not be interested :D

PR #267, opened 2023-01-28
Deprecating current support of `showyourwork` command line with arguments and no subcommand

dfm:
Currently you can run `showyourwork` directly as an alias for `showyourwork build`. Then any extra arguments that are unrecognized by `showyourwork` are passed directly to `snakemake`. This is useful for cases like:

```bash
showyourwork
```

or

```bash
showyourwork --rerun-incomplete
```

but it can be bad in cases like (e.g.):

```bash
showyourwork arxiv
```

which runs without error, silently ignoring the argument. This is a problem for people like me who don't know all the valid subcommands off the top of their head.

This PR deprecates this behavior except in the specific case of `showyourwork` with no arguments at all. There are no functional changes introduced, except that a `FutureWarning` is emitted in the case of deprecated use. This might result in users being required to type `build` more often (I always do anyways…), but I think that's worth the cost given the opportunity to avoid confusing behavior in other cases. We can discuss better handling logic in future PRs.

### DIFF
--- a/showyourwork/cli/__init__.py
+++ b/showyourwork/cli/__init__.py
@@ -11,6 +11,7 @@ article.
 
 """
 import sys
+import warnings
 
 from .. import __version__
 from .main import DEFAULT_SUBCOMMAND, OPTIONS, SUBCOMMANDS, main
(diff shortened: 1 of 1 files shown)


dfm:
cc @rodluger, @katiebreivik, @jfcrenshaw, @mathren, @arm61, @kazewong

As users of `showyourwork`, any concerns about deprecating this behavior in the next major release?

arm61:
My only comment would be that my expectation for running `showyourwork` without any arguments would be for the software to do nothing and return a `--help` return. Similar to if you run `git` from the command line. 

I just think that unless I explicitly want to `build` (which could be a very long task, with `snakemake`-shaped pitfalls if you cancel the run at the wrong time), it is safer to do nothing.

dfm:
@arm61 — I totally agree! The existing behavior is that `showyourwork` is (approximately) an alias for `showyourwork build`, which can cause some nasty issues, so I'm definitely in favor of deprecating. I'd be even happier to do what you suggest and just print the help info when no subcommand is provided!

arm61:
If this is something that a newbie developer could contribute, I am happy to have a bash myself and save you the effort (if you just point me in the right direction).

dfm:
Absolutely! It would require a change here (we could probably even just remove the `entry_point` function, and just use `showyourwork.cli.main`):

https://github.com/showyourwork/showyourwork/blob/62a530eb94e3514429d2abe902b20524b263bd3e/showyourwork/cli/__init__.py#L20

and here (to provide sensible behavior when no arguments are included):

https://github.com/showyourwork/showyourwork/blob/62a530eb94e3514429d2abe902b20524b263bd3e/showyourwork/cli/main.py#L76-L79

I don't think we want to have this implemented before the next major release, so there isn't a huge rush, but please do open a PR if you're interested!

arm61:
I will have a bash at tomorrow!

katiebreivik:
Agree with @arm61 for longer term response re:--help, but am happy for the deprecation given that this issue has also caused confusion in my own workflow a couple of times.

dfm:
Thanks @katiebreivik!

I've merged this PR and I'll plan on having this deprecation in the next minor release. @arm61 has already started working on the PR to remove this support in the next major release!

PR #272, opened 2023-01-30
No arguments will give help information 

arm61:
I have chosen to stick with the `entry_point` approach, removing this raises some issues with the tests failing that are beyond my current understanding of how `showyourwork` operates.


dfm:
I actually think the approach in #271 was better because I'd prefer to depend on `click`'s handling of the logic rather than re-implementing it ourselves. The test errors were appearing only because of a missing import - easily fixable! How would you feel about re-opening that one and giving it another shot?

arm61:
Can do!

arm61:
Okay, it looks like that works now. Thoughts?

dfm:
I don't know what was up with the pre-commit check, but looks like we've got it working now!

So simplify the diff, I renamed `__init__.py` back to `main.py` then imported `main as entry_point` in `__init__`. It's really just a matter of taste, but I prefer to have as little as possible in the `__init__`. Otherwise I think this is good to go as soon as we get all green on the tests! Thank you 🚀 

My plan is to merge this into the `main` branch, and we'll use the `main` branch to target `v0.5`. I've created a `v0.4.x` branch to track minor releases on the `v0.4` trajectory.

PR #278, opened 2023-02-01
Add a reminder to clear cache when static figures are updated

matiscke:
When a static figure is changed, the showyourwork build on github does not take the update into account right away because it uses a cached version. This can lead to the situation where an old figure ends up in the PDF even though the updated version is in the `src/static/` directory.

I suggest to add a line in the docs to spare inexperienced users like myself the troubleshooting.


dfm:
Interesting - thank you!!

I'm inclined to try to fix this in showyourwork instead of documenting this failure 🙈... I have an idea about where it might be coming from - let me take a quick look!

matiscke:
Since #279 fixes this issue, I think this PR can be closed.

dfm:
Thanks for reporting the issue @matiscke!

PR #279, opened 2023-02-01
Fixing #278 by adding explicit dependencies on static figures

dfm:
@matiscke — This should fix the issue you were seeing in #278. The basic problem is that we weren't explicitly tracking the figure in the `static` directory as a dependency of the generated figure in `figures`. This meant that when determining whether or not to update the copy `showyourwork` wouldn't take into account the modification time of the  static figure. This was also an issue locally, not just on GitHub Actions, and I've confirmed that this fix does the trick for me!


dfm:
~~For reasons that I don't understand, this is breaking the remote unit tests. I need to investigate.~~ Figured it out!

matiscke:
Excellent, thank you for fixing this @dfm !

PR #285, opened 2023-02-11
Major rewrite of the core infrastructure

dfm:
This is the beginnings of a major rewrite of the `showyourwork` back end that I've been working on for the last little while. My plan is to get this to the point where the user experience won't be very different. The configuration won't be 100% backwards compatible, but it should be pretty close.

The main user-facing difference will be that the figure dependencies will no longer be specified using the `\script` command in the TeX manuscript. Instead, the data dependency structure will be completely inferred from the `snakemake` DAG. To get `showyourwork` to automatically generate rules for your figures or other dynamically generated files, you'll need to instead specify these dependencies using the following syntax in the config file:

```yaml
dynamic:
  - script: path/to/script.py
    output: path/to/figure.pdf
```

where `output` can also be a list of output files. This change in syntax is nice for a few reasons, but I big one is that it means that the dependencies labelled in the document will always be _exactly the dependencies required to generate the file_ (because `snakemake` knows that already without us needing to specify it!), even without the `dependencies:` entry in the config file (which no longer exists).

From the back end perspective this rewrite is motivated by the following:

1. I wanted to be able to support more flexible workflows. For example, I didn't want to require the specific `showyourwork` directory structure. Rewriting makes this easier because expectations about the file system were strongly baked into the previous implementation. Other workflows include future developments like support for mystjs-based build systems.

2. As part of supporting these flexible workflows, I wanted to make `showyourwork` more flexible and extensible. This requires well-documented (to-do) hooks to allow customization. This was hard with the existing implementation in part because of the fact that `snakemake` was executed twice for every build. The change above ☝️ means that we only need to fire `snakemake` up once, and makes it somewhat easier to provide hooks. This should also make everything faster!

3. I wanted to improve the testing infrastructure, and in particular add unit tests for the core functionality that didn't require the expensive tests that generate new GitHub repositories that currently provide the majority of test coverage.


dfm:
I've mostly added support for Windows, but the dependency tree solver doesn't seem to work on Windows for reasons that I don't totally understand. It would be good to try to debug it on a Windows machine instead of on GitHub Actions.

PR #308, opened 2023-02-21
Enable and document monochrome and custom color margin icons

### LINKED ISSUE #290: Monochrome margin icons
For some the coloured margin icons are a bit garish. It would be nice to have an option in the `yml` to customise these colours (or specifically make them black).

The colours seem to be hardcoded here: https://github.com/showyourwork/showyourwork/blob/697779a1c7f787fe9e1a3ff15fcdd86649ddf845/showyourwork/workflow/resources/styles/build.tex#L58
Tips on how the stuff defined in the `showyourwork.yml` gets into there?

arm61:
This PR will close #290. 

It allows the margin icons to be forced to be monochrome, with the boolean `margin_icons.monochrome` (this will override custom colours). 

If the users would like custom colours these can be set as rgb strings for `margin_icons.colors.github` (or `sandbox`/`dataset`/`cache`). 

The decision to have `monochrome` the level above has two benefits: 
- Less yaml for the user
- Cleaner implementation of the loop in [compile_setup.py](https://github.com/showyourwork/showyourwork/compare/main...arm61:issue-290?expand=1#diff-aef33a3097c39d53a3da97ac621516421407a54b08b0a9fb6410111a32980828)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -522,7 +522,7 @@ Finally, dependencies of the manuscript file are also allowed:
 
 **Type:** ``bool``
 
-**Description:** Makes all margin_icons black, this will override and custom colors.
+**Description:** Makes all margin_icons black, this will override any custom colors.
 
 **Required:** no
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -441,6 +441,96 @@ Finally, dependencies of the manuscript file are also allowed:
     src/tex/ms.tex:
         - src/tex/stylesheet.tex
 
+.. _config.margin_icons:
+
+``margin_icons``
+^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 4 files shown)


docs/config.rst
```diff
@@ -441,6 +441,96 @@ Finally, dependencies of the manuscript file are also allowed:
...
+**Required:** no
+
+.. _config.margin_icons.monochrome:
+
+``margin_icons.monochrome``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**Type:** ``bool``
+
+**Description:** Makes all margin_icons black, this will override and custom colors.
```
arm61:
```suggestion
**Description:** Makes all margin_icons black, this will override any custom colors.
```

Issue #307, opened 2023-02-21
Enhance intro docs to clarify relationship between Snakemake and SYW!


maxisi:
Working on adding to SYW! intro docs to make it more transparent to users who haven't encountered Snakemake (or the notion of workflows) before. (@maxisi working on this during hackathon)

dfm:
Closed by #311

PR #310, opened 2023-02-21
Snakemake howto

katiebreivik:
This is a borderline ridiculously simple PR that seeks to make the Snakemake use in SYW a bit more obvious and the introduction a bit more gentle. The previous version of the docs had basically the same information, but just in a different order.

Happy to make any wording modifications that folks would like to suggest!

Should help with closing #307


docs/conf.py
```diff
@@ -1,3 +1,7 @@
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath("."))
```
adrn: Suspicious of this! We're debugging in Real Life™, so let's keep working on that.

dfm: I'd probably vote for removing this from here, and opening it in a separate PR, perhaps also addressing https://github.com/showyourwork/showyourwork/issues/312 from @rwegener2. What do you think of that?

katiebreivik: I think that makes sense! Re-commited conf.py below; I should be able to dig a bit more on the hacky import/docs build PR later today.

dfm:
🎉🎉

Thanks @katiebreivik!!

docs/conf.py
adrn: I have a theory here! I predict that @katiebreivik and @afeinstein20 have edited their `$PYTHONPATH`'s by setting them like `export PYTHONPATH=/some/path` in their shell rc or profile file. The `$PYTHONPATH` has to have a blank entry in it for Python to include the current directory in the search path (i.e. in `sys.path`). 

I modify my `$PYTHONPATH` locally, but I do it like:
```bash
export PYTHONPATH=$PYTHONPATH:/some/path
```
The default `$PYTHONPATH` is a blank space, so this is equivalent to:
```bash
export PYTHONPATH=:/some/path
```

With this, if I do
```bash
python -c "import sys; print(sys.path)"
```
in a path, I see that path in the list - e.g., in my home dir:
```
['', '/Users/apricewhelan', '/Users/apricewhelan/anaconda3/lib/python310.zip', ... more crap ...]
```

But without the empty entry in the `$PYTHONPATH`, my home dir disappears from the list.

katiebreivik: So! interestingly, I haven't actually modified the `$PYTHONPATH`, _but_ when I do
`python -c "import sys; print(sys.path)"`
I don't have the current directory on my path; so regardless of the directory, I get:
`['', '/Users/kbreivik/opt/anaconda3/envs/cosmic/lib/python310.zip', ...]`
So I have the empty entry, but not the current directory.... 

@afeinstein20 : is this also the case for you?

PR #314, opened 2023-02-21
Fast draft mode edit

kazewong:
Here is a checklist of desire features in draft mode

- [x] Draft mode options in showyourwork cli
- [x] Fast compilation with runtime similar to just doing pdflatex with only text
- [x] Correct citations 
- [ ] Move the output to the main directory
- [ ] Use static figures if they exist. If they don't, use a place holder image


HealthyPear:
Hi @kazewong  could you try to rebase this branch on top of the current main and force push?

It should be almost flawless, perhaps you will have to update only the `showyourwork/cli/commands/__init__.py` file adding `from .draft import draft` to the version in `main` as the rest of the modifications are either additions or unrelated modifications (hopefully).

PR #319, opened 2023-02-22
Add section to docs describing how to setup docs env

rwegener2:
### What was done and how
This PR adds instructions in the Developer section describing 1) how to setup the environment for building the docs and 2) how to run the docs locally. It seems that the developer docs right now consist of just the API reference, so this section was added on the main page of the Developer section.

### A related idea
Something else that could improve the docs contribution experience is if github was configured to build the docs whenever a PR gets made. That makes it easier to check that the formatting is rendering as desired. If that is something folks want it looks like something an admin can setup, [described here](https://docs.readthedocs.io/en/stable/guides/pull-requests.html).

### Not addressed in this PR
The conversation happening [over here](https://github.com/showyourwork/showyourwork/pull/310#discussion_r1114296054) about how to deal with the `import hacks` error. This PR concludes what I worked on yesterday, although I'm not sure that I'll have more time this week to pursue the import error. If a solution is found folks are super welcome to add that to this PR, or I can quickly make a commit. Either way, I'll keep this PR as a draft until we decide if fixing the import should get added to this PR or done in a different one.


dfm:
Thanks! Some quick comments:

> Something else that could improve the docs contribution experience is if github was configured to build the docs whenever a PR gets made. That makes it easier to check that the formatting is rendering as desired. If that is something folks want it looks like something an admin can setup, [described here](https://docs.readthedocs.io/en/stable/guides/pull-requests.html).

This [was already configured](https://readthedocs.org/projects/showyourwork/builds/), and [here's the version of the docs for this PR](https://showyourwork--319.org.readthedocs.build/en/319/). There is some sort of permissions issue meaning that ReadTheDocs couldn't report this link back here, but I may have fixed that ... we'll see!

Also, what do you think about adding this as a separate page in the main documentation index called something like "Contributor guide"? Then we could continue to expand it with other advice about contributing. I feel like it's a bit hidden in the "Developer API" section right now, and it might be nice to highlight it more explicitly!

rwegener2:
Sorry I fell off the radar, there. Thanks so much for the feedback @dfm! I think a separate contributor page is a great idea.

The latest commit creates the new section moves the docs build section to that new page. I added a small section called "open issues", but I'm pretty unopposed if we just want to keep that out. It basically just points people to the Github Issues page.

dfm:
Thanks @rwegener2! This looks perfect to me. Let's merge and we can edit in different PRs if we want. Thanks again for adding this!!

dfm (approved):

PR #324, opened 2023-02-24
Allow horizontal moving of margin icons

arm61:
Sometimes the margins icons were closer to the other column than the one they are associated with. This (slightly hacky) fix allows the user to control the margin icons horizontal offset.


dfm:
Thanks @arm61! I don't _love_ moving that whole block over to the `metadata` file. Do you think it would be possible to define a macro in `metadata` and then use some sort of `ifdef` in the main `showyourwork.tex` file? I'm not sure exactly what the syntax would look like...

arm61:
Oft, that sounds like tex-fo beyond my current skills. I will give it some thought though.

dfm:
OK sounds good. I don't feel extremely strongly about this, but I'll also have a think. If we can't figure it out, I'm happy to merge this as is.

arm61:
One way might be to see if we can add just the single line in the correct place on the dynamically build `.tex`?

dfm:
Yes - good idea! Maybe we should just get rid of the `metadata` file and instead just template the full `build.tex`. What do you think of that?

I think the approach I would take would be to rename `build.tex` to `build.template.tex` or something, and then in `compile_setup.py` we would load that in instead of defining the template as an inline string.

arm61:
How do we feel about merging this as the alternate (which I like) seems like a much bigger job?

dfm:
Absolutely! I'll do that now. Sorry about the delay.

arm61:
No worries about the delay. I sat on it a bit to think if it was a worthwhile change.

PR #333, opened 2023-03-07
Add instructions about using mamba

### LINKED ISSUE #330: Document the use of `mamba` for solving environments
I see that this already popped out in issues like #240, but I felt the need to ask/suggest explicitly.

Ae already using mamba to create the conda envs?  If not I'd suggest to do it as it is sensibly faster.
You could use the mambaforge distribution.

HealthyPear:
Closes #330


dfm:
Thanks!! 🚀

PR #342, opened 2023-04-15
Update VSCode defaults with preferred build command

MilesCranmer:
`showyourwork` was deprecated in favor of `showyourwork build`. This updates the default VSCode settings.


MilesCranmer:
The current version of LaTeX workshop seems to need the `--project` setting from #348 , as it wants to run the build command in the `src/tex` folder. Thus I'm closing this PR until a general solution can be found.

PR #348, opened 2023-04-28
Enable custom root directory of build commands

### LINKED ISSUE #347: Option to specify build directory
Right now you can only run `showyourwork build` in the root directory of a showyourwork project. This means that whenever you are running syw from a different directory, you need to manually change directories before building.

What makes this more difficult is that some tools, like LaTeX-Workshop, make it impossible to run a sequence of shell commands. You have to specify a "command" and then a sequence of arguments. Thus in some scenarios this creates friction in getting syw to build unless you get lucky and have the default directory = syw root directory.

A solution to this would be to add a `-d, --directory <SYW_PROJECT_ROOT>` flag to all showyourwork commands. This is standard for `make`-like tools so I think it would be a good addition. 

Not sure what a good option name would be, it seems like there isn't a really standard:

- `make` and `poetry` use `-C, --directory`
- `snakemake` uses `-d, --directory`
- `pip` uses `--root`
- `cmake` takes the path as argument, and also allows `-S` for source/`-B` for build.
- `pdflatex` and `latexmk` just take the source as argument.

So maybe we just go with `-d, --directory` to keep consistency with snakemake? @dfm?

MilesCranmer:
Edit: This is lower priority than #350 and #351 

---

This adds the `-p, --project <root folder>` CLI option so that you can run `showyourwork` from other directories.

I didn't want to break anything so the change is extremely simple: it just changes directories, runs the regular showyourwork commands, then changes back.

@dfm would you be willing to review this?

Fixes #347

### DIFF
--- a/showyourwork/cli/main.py
+++ b/showyourwork/cli/main.py
@@ -1,6 +1,7 @@
 import os
 import re
 import shutil
+from collections import namedtuple
 from contextlib import contextmanager
 from textwrap import TextWrapper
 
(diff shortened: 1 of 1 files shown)


dfm:
I'm up for including something like this, but I don't love this specific implementation since you'll end up in the wrong directory if anything goes wrong during the `showyourwork` run. I'd recommend something like the following context manager:

```python
import os
from contextlib import contextmanager

@contextmanager
def in_project(path):
    old_dir = os.getcwd()
    os.chdir(path)
    yield
    os.chdir(old_dir)
```

which you could then use as

```python
with in_project(project):
    ensure_top_level()
    commands....
```

to make sure that we always end up chdiring back.

MilesCranmer:
Okay I added the context manager, great idea!

~~Also, I refactored the click options which are used multiple times so there wasn't as much repetition. For example:~~

```diff
- @click.option(
-     "-c",
-     "--cores",
-     default="1",
-     help="Number of cores to use; passed to snakemake.",
- )
+ @takes_cores
```

~~Let me know what you think.~~

MilesCranmer:
Sorry, am being a perfectionist... I refactored the common options again to be like this instead:

```py
@click.option(*cores_option.args, **cores_options.kwargs)
@click.option(*conda_frontend_option.args, **conda_frontend_option.kwargs)
```

etc. So at least with this it still is obvious that it's a `click.option` rather than some crazy internal macro.

dfm:
LGTM - thank you!

MilesCranmer:
Thanks for merging!

Do you think this, #350, and #351, could all be released in a new version of SYW? Then I can give a tag instead of the SHA.

dfm:
Will do!

PR #350, opened 2023-04-28
Fix missing tectonic args

### LINKED ISSUE #349: `config['user_args']` was removed by #266
@dfm the PR #266 deleted some of the config options. I think this was a mistake because the config options are still documented and parsed from the config file, and are necessary for configuring tectonic.

Before the refactor, in the file `tex.py`, there were some extra config options:

https://github.com/showyourwork/showyourwork/blob/a3d8b337ec1390eb021e63377643f79bc8d01783/showyourwork/tex.py#L33-L40

https://github.com/showyourwork/showyourwork/blob/a3d8b337ec1390eb021e63377643f79bc8d01783/showyourwork/tex.py#L62

However, after the refactor, only the contents of `force_args` remains:

https://github.com/showyourwork/showyourwork/blob/4dd722a5f9474a9de656573aebdebaa3fc91ecd9/showyourwork/workflow/rules/compile.smk#L61-L66

So any options I pass for `tectonic_args` (which is loaded to `'user_args'`) are simply dropped.

I think #317 raised an issue about some of these missing config options, but other ones are still missing...

Was this a planned deprecation or should we fix it? :smile:

MilesCranmer:
Fixes #349

@dfm could you review?

There are likely some other missing config options but I'd prefer this be merged quickly because my paper is coming out soon and the SYW version requires `shell-escape` to build :)

### DIFF
--- a/showyourwork/workflow/rules/compile.smk
+++ b/showyourwork/workflow/rules/compile.smk
@@ -7,8 +7,6 @@ Runs the script :doc:`pdf` to generate the article ``ms.pdf``.
 from pathlib import Path
 from showyourwork import paths
 
-maybe_synctex = "--synctex" if config["synctex"] else ""
-joined_user_args = " ".join(config["user_args"])
 
 rule:
     """
@@ -58,15 +56,19 @@ rule:
         (paths.user().compile / f'{config["ms_name"]}.synctex.gz').as_posix(),
     conda:
         (paths.showyourwork().envs / "tectonic.yml").as_posix()
+    params:
+        maybe_synctex="--synctex" if config["synctex"] else "",
+        user_args=" ".join(config["user_args"])
     shell:
         """
         cd "{input.compile_dir}"
-        tectonic                  \\
-            --chatter minimal     \\
-            --keep-logs           \\
-            --keep-intermediates  \\
-            {maybe_synctex}       \\
-            "{input[0]}" {joined_user_args}
+        tectonic                      \\
+            --chatter minimal         \\
+            --keep-logs               \\
+            --keep-intermediates      \\
+            {params.maybe_synctex}    \\
+            {params.user_args}        \\
+            "{input[0]}" 
         """
 
 # TODO: Add config options for verbosity?
--- a/showyourwork/workflow/rules/compile.smk
+++ b/showyourwork/workflow/rules/compile.smk
@@ -7,7 +7,6 @@ Runs the script :doc:`pdf` to generate the article ``ms.pdf``.
 from pathlib import Path
 from showyourwork import paths
 
-
 rule:
     """
     Setup the temporary files for compilation.
--- a/showyourwork/workflow/rules/compile.smk
+++ b/showyourwork/workflow/rules/compile.smk
@@ -72,7 +72,7 @@ rule:
             --keep-intermediates      \\
             {params.maybe_synctex}    \\
             {params.user_args}        \\
-            "{input[0]}" 
+            "{input[0]}"
         """
 
 # TODO: Add config options for verbosity?
--- a/showyourwork/workflow/rules/compile.smk
+++ b/showyourwork/workflow/rules/compile.smk
@@ -7,6 +7,9 @@ Runs the script :doc:`pdf` to generate the article ``ms.pdf``.
 from pathlib import Path
 from showyourwork import paths
 
+maybe_synctex = "--synctex" if config["synctex"] else ""
+joined_user_args = " ".join(config["user_args"])
+
 rule:
     """
     Setup the temporary files for compilation.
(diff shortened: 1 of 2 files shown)


showyourwork/workflow/rules/compile.smk
```diff
@@ -7,6 +7,9 @@ Runs the script :doc:`pdf` to generate the article ``ms.pdf``.
 from pathlib import Path
 from showyourwork import paths
 
+maybe_synctex = "--synctex" if config["synctex"] else ""
+joined_user_args = " ".join(config["user_args"])
```
dfm: It's best practice to move these to the `params` section of the rule below. For example,

```python
params:
    maybe_synctex="--synctex" if config["synctex"] else "",
    user_args=config["user_args"]
```

should do the trick (you shoulndn't need to join them manually).

showyourwork/workflow/rules/compile.smk
```diff
@@ -62,10 +65,13 @@ rule:
             --chatter minimal     \\
             --keep-logs           \\
             --keep-intermediates  \\
-            --synctex             \\
-            "{input[0]}"
+            {maybe_synctex}       \\
+            "{input[0]}" {joined_user_args}
```
dfm: Then here I don't think this will work as currently written because the rule won't have access to the globals. Instead, this should be:

```suggestion
            {params.maybe_synctex}   \\
            {params.user_args}             \\
            "{input[0]}"
```

showyourwork/workflow/rules/compile.smk
```diff
@@ -62,10 +65,13 @@ rule:
...
         """
 
+# TODO: Add config options for verbosity?
+# See config.py and old tex.py for missing configs.
```
dfm, line 79: Perhaps open this as an issue so we can remember?

showyourwork/workflow/rules/preprocess.smk
```diff
@@ -8,6 +8,7 @@ file containing metadata about the build and the workflow graph.
 """
 from showyourwork import paths
 
+joined_user_args = " ".join(config["user_args"] )
```
dfm: As in `compile`, this will need to be in `params`.

showyourwork/workflow/rules/preprocess.smk
```diff
@@ -56,7 +58,7 @@ rule:
             --keep-logs           \\
             --keep-intermediates  \\
             -r 0                  \\
-            "{input[0]}"
+            "{input[0]}" {joined_user_args}
```
dfm: User args should be before `input`.

dfm (changes requested):
Thanks! I'm happy to merge this, but I don't expect that it currently works as written, because snakemake rule substitution doesn't typically have access to the globals. Can you test and confirm that after the suggested changes this works for your use case?

MilesCranmer:
It has been working for me so far (have been using a local copy). The only way I can compile my paper which uses `minted` is to get the `-Z shell-escape` argument through to tectonic.

showyourwork/workflow/rules/preprocess.smk
MilesCranmer: I thought so too! But in tectonic they actually go after

showyourwork/workflow/rules/compile.smk
MilesCranmer: Cool. Thanks!

showyourwork/workflow/rules/preprocess.smk
MilesCranmer: It's weird. I'm doing a sanity check right now and looks like there are examples of both in the docs and in user issues... And when I initially tried putting the `-Z shell-escape` before the input, it failed. Very strange.

dfm: The old behavior had them first: https://github.com/showyourwork/showyourwork/blob/a3d8b337ec1390eb021e63377643f79bc8d01783/showyourwork/tex.py#L62

dfm: I'm about 99% sure that the syntax as written isn't adding any arguments, so I'm surprised you're finding it to work at all!

MilesCranmer: Sometimes the 1% wins...
image

image

Very weird!

MilesCranmer: Anyways I will definitely change it, not to worry. I'm just marveling at why it works...

MilesCranmer: For posterity:

image

So looks like tectonic treats its arguments as an unordered set...

dfm: The 99% was about the snakemake access to globals, I'm never confident about anything tectonic related :). I'm worried that you're seeing it working because of a caching issue rather than the new command working, but I could be wrong.

MilesCranmer: I did a `showyourwork clean` in between so that's probably it!

MilesCranmer:
Okay it should be ready. Just confirmed that this branch still builds the paper for me. The issue I saw earlier with the out-of-order user args must have been a dumb mistake I made (which uncovered tectonic weirdness).

dfm (approved):

dfm:
LGTM - thank you!!

PR #351, opened 2023-04-28
Allow user-defined tectonic.yml

MilesCranmer:
This is a really simple change that checks the user's repository for a `tectonic.yml` file. If not found, it falls back to the repository version.

This is useful if you need to use a different teconic version, or, in my case, if you need an extra graphics library/fontspec/etc for your compilation (since the build environment is completely isolated).

In my case I needed the `pygments` library which is used by the LaTeX package `minted` - something I can't do outside of the LaTeX process - so it needs to be added to the tectonic env. So, I simply created a custom `tectonic.yml` and added this `pygments` library to it, and now everything works.

---

@dfm sorry for the spam but do you think you could review and merge?

I can add docs on this later, I just need to get this working quickly so my paper coming out later today actually compiles without hacks, since it's a nice demo of what SYW can do :P

### DIFF
--- a/showyourwork/workflow/rules/compile.smk
+++ b/showyourwork/workflow/rules/compile.smk
@@ -7,6 +7,11 @@ Runs the script :doc:`pdf` to generate the article ``ms.pdf``.
 from pathlib import Path
 from showyourwork import paths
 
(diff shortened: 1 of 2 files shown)


dfm:
LGTM - thanks!

PR #352, opened 2023-04-28
Add `snakefmt` to pre-commit

MilesCranmer:
Seems like a good addition given the number of snakemake files and the fact that `black` is already used for Python:

> This repository provides formatting for [Snakemake](https://snakemake.readthedocs.io/) files. It follows the design and specifications of [Black](https://black.readthedocs.io/en/stable/).


dfm:
This looks great too! Can you fix the conflicts and try to see what the pre-commit failure is being caused by? Here is the error log:

````
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
isort....................................................................Passed
black-jupyter............................................................Passed
pycln....................................................................Passed
snakefmt.................................................................Failed
- hook id: snakefmt
- exit code: 1

snakefmt.exceptions.InvalidPython: Black error:
```
Cannot parse: 89:0: else:
```
````

Seems to be a syntax error somewhere, but it's not very clear :D

**Comment by MilesCranmer** [comment]
This syntax error happens on master as well. It's due to `build.smk` on this line:

https://github.com/showyourwork/showyourwork/blob/8038af82f9bfcb42d451760b3cf49585b9a3d3ad/showyourwork/workflow/build.smk#L89

This is at the limits of my snakemake knowledge as I didn't even know you could have rules within if statements. Any idea?

**Comment by MilesCranmer** [comment]
Nevermind I fixed it. Snakefiles are a little too expressive for my taste; it was some issue where it couldn't tell where a block ended and the next one began, so I put a `pass` at one indent down so the formatter could figure out it had to exit the scope.

File: showyourwork/workflow/build.smk [hunk 1]
```diff
@@ -18,44 +22,39 @@ workdir: paths.user().repo.as_posix()
 # What kind of run is this? (clean, build, etc.)
 run_type = get_run_type()
 
+if not (paths.user().temp / "config.json").exists() and run_type != "clean":
+    raise exceptions.MissingConfigFile()
```
MilesCranmer, line 26: I just made a second if statement instead of an else. But I'm not sure if `onsuccess` refers to the original if statement or not?

Also - with this exception raised at the top, does the rest even need to be embedded in an if statement?

showyourwork/workflow/build.smk
```diff
@@ -18,44 +22,39 @@ workdir: paths.user().repo.as_posix()
...
     report: "report/workflow.rst"
 
-
     # Remove old flags
     for file in paths.user().flags.glob("*"):
         file.unlink()
 
-
-    # Set up custom logging for Snakemake
+        # Set up custom logging for Snakemake
```
dfm, line 42: This formatting is wrong. Are there settings that we can change in snakefmt?

showyourwork/workflow/build.smk
```diff
@@ -69,37 +68,29 @@ if (paths.user().temp / "config.json").exists():
...
     if config["require_inputs"]:
         patch_snakemake_missing_input_leniency()
 
-
-else:
-
-
-    if run_type != "clean":
-        raise exceptions.MissingConfigFile()
+    pass
```
dfm, line 85: After adding the new conditional at the top, this isn't still needed anymore, is it?

dfm (changes requested):
I added some tiny inline comments, but this otherwise looks great!

MilesCranmer:
It looks like this might be a bug in snakefmt: https://github.com/snakemake/snakefmt/issues/190 and https://github.com/snakemake/snakefmt/issues/186

We could wait to merge this until those bugs are fixed, lest it produces a bug on our side due to incorrect formatting. Wdyt?

showyourwork/workflow/build.smk
MilesCranmer: I think you are correct.

Actually, could the entire if statement be removed now, since the check is at the top?

dfm:
Sounds like a plan to me!

HealthyPear:
The 2 upstream issues mentioned here have been closed / fixed, so I think we should revive this PR!

HealthyPear:
I think the conflicts could be a bit tedious to fix here, but I am not the author of the PR, so I will leave the decision to @MilesCranmer if he wants to try and update his branch or we open a new PR with the same basic change (I guess only `.pre-commit-config.yaml`, as the rest of the changes - that now trigger the conflicts - should be a consequence of the commit introduced by pre-commit).

The only relevant commit should be `c5a139378d5bfc02d104028519050ea4af99188b` so maybe you could even rebase this branch by dropping all commits after that, force-push to the same branch and rebase from main using the GUI, so pre-commit will commit on the other files pointing to main?

MilesCranmer:
Yes I think it needs to basically be done on a new branch. But it's pretty minor so I'll close it for now

HealthyPear:
OK! let me know if you will do it or you want me to, I think it's a good idea to have it!

PR #354, opened 2023-04-29
New overview figure

### LINKED ISSUE #343: Update overview figure to include other language scripts
The overview figure has the "Python" logo next to the "pipeline scripts" and "figure scripts":

image

I think this is misleading for people checking out the project because they may immediately think their {Julia/R/Rust} scripts can not be used in their SYW project. (In fact, the reason I am aware of this issue is because a Julia person thought they couldn't use SYW). However, Snakemake has native support for all four: https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#external-scripts, and is also able to execute arbitrary shell commands.

@rodluger would you happen to have the source file for this overview figure? I'd propose that we either tweak it to have other language logos, or just remove any particular language's logo altogether. What do you think @dfm?

---

side note: it could be useful to include `paths.jl`, `paths.rs`, `paths.R` in the cookie-cutter template. Maybe we could have a "languages to support" option when creating the template repository that would download the needed scripts in each case, and update the `showyourwork.yml` file to map the filetypes.

MilesCranmer:
Fixes #343 

~~Here's the adjusted image~~ (edit: I changed it again - see below)

image

I note that I just pasted figures on the original png as I don't have the original vector artwork (@rodluger if you see this and happen to still have that figure, please share!).

The languages I included are the ones that snakemake supports out of the box, but hopefully it should be clear from the number of languages that showyourwork is language independent.

I also created a new internal repository https://github.com/showyourwork/showyourwork-artifacts to store artifacts so they won't pollute the main repository's history/increase download time. I added the .psd file so it can be edited in the `overview_v2` branch ([photopea.com](https://www.photopea.com/) is what I used).


MilesCranmer:
I didn't like how overlapped the icons were so I changed it again:



To get this I basically took one of those paper icons from the "environment" column, cropped the top-left corner, rotated it around, and stuck it over top of the original python...

dfm:
The updated figure looks good to me! I think it would be better to just copy the figure over to this repo once we're happy though, rather than relying on an external link. That would be a somewhat more fragile approach!

MilesCranmer:
The URL references the specific SHA of the committed image, and the host repo is within the `showyourwork` organization, so I think (?) it's safe? The source file is 4 MB and overview figure is 400 kB so I think it's just better practice to keep a separate docs/artifacts repo so `git clone` of the source code doesn't get too unweildy, far into the future, as the docs grow and grow. Plus it's just an image so if the link breaks for some reason, no code will be hurt in the process. Wdyt?

MilesCranmer:
By the way, any idea what's going wrong with that integration test? The docs seem to be building fine...

dfm:
I'd prefer to copy it over! That way the docs are a self-contained entity. I don't expect that we'll change it frequently so I'm not concerned about git bloat.

(Don't worry about the integration test, it's [a known issue (without a known solution)](https://github.com/conda-forge/tectonic-feedstock/issues/59) and I always just re-run it by hand when it fails...)

MilesCranmer:
Ok, fine with me. Just pushed.

(I'll leave the `.psd` source file on the artifacts repo as it is 4MB and isn't needed to render the docs; it's just if someone wants to edit the photo)

dfm:
Perfect - thank you!!

dfm:
🚀

PR #364, opened 2023-05-02
Enable preprocessing of arXiv submission contents

### LINKED ISSUE #356: Post-processing rules for arXiv tarball?
Is there an expected way to write post-processing rules for generating the arXiv tarball? Maybe a way I can hack the specific Snakemake rule that is being triggered?

Basically I would like to run a few commands whenever the arXiv tarball is generated:

1. Run `latexpand` to remove comments and `de-macro` to expand all macros. Both of these make it easier for scraping tools like Google to index your paper, as well as humans to read the source.
2. Store my `pygment` cache and convert `\usepackage{minted}` to `\usepackage[frozen]{minted}`, which makes it compatible with arXiv where you can't use `shell-escape`.

Perhaps rather than having these implemented internally it would be great if I could just write some bash commands at the arXiv tarball stage. There's probably other packages too that require custom post-processing before putting up on arXiv.

(Any idea if this is possible @dfm?)

MilesCranmer:
Fixes #356.

This is a really simple modification to allow user processing of arXiv tarballs.

It introduces a new user config of the form:

```yaml
preprocess_arxiv_script: "my_preprocessing_script.sh"
```

This script will be executed **during** the `syw__arxiv` step, and be passed the directory of the expanded arXiv tarball. The script will be allowed to modify the tarball contents in-place.

I tried a Snakemake strategy as discussed in #356 but it just seemed like it would make things way more complicated for the user. So, despite this strategy being a bit heavy-handed, it's also simple, robust, and easy for the user to configure.

@dfm wdyt?


dfm:
I don't love this, but I do think it's a good compromise within the current constraints! Can you add a test?

MilesCranmer:
For unit-testing it doesn't look like there are any tarball tests at all so it might be a bit of work to add that. If needed I think I'll have to come back to this in a few weeks after neurips deadline and maybe until then I can just reference my git branch in the action.

It should basically be a no-op for existing users though, aside from the extra copy. (I'm testing it locally and seems to work.)

MilesCranmer:
Hey @dfm, would it be possible to have this merged? I realized that the current state of my PySR paper still requires my branch specific changes for building, as it needs the pygment fix from this PR.
Thanks!
Miles

MilesCranmer:
Here's an example preprocessing script if this helps clear things up (I could put this on the docs too)

https://github.com/MilesCranmer/pysr_paper/blob/main/src/scripts/preprocess_arxiv_tex.sh

```bash
#!/bin/bash
pushd $1
sed -i ms.tex -e 's/finalizecache/frozencache/g'
sed -i ms.tex -e '/PassOptionsToPackage/d'
sed -i showyourwork.tex -e 's/RequirePackage{xcolor}/RequirePackage[table]{xcolor}/g'
popd
```

We could also have it automatically change to the arXiv tarball directory instead of passing it as an argument.

showyourwork/workflow/scripts/arxiv.py
```diff
@@ -26,8 +30,22 @@
...
-            if file.name not in exclude and file.suffix != ".bib":
-                tarball.add(file, arcname=file.name)
+    with TemporaryDirectory() as tmpdir:
+        # First, copy to a temporary directory, excluding files
+        shutil.copytree(paths.user().compile, tmpdir, dirs_exist_ok=True)
+
+        if config["preprocess_arxiv"]["enabled"]:
+            # Run the preprocessing script, if provided:
+            script = config["preprocess_arxiv"]["script"]
+            if not script:
```
dfm: Why do we need the `enabled` config if the real requirement is that the `script` be provided. Why not just have a config called `preprocess_arxiv`, which is just the name of the script?

dfm (changes requested):

showyourwork/workflow/scripts/arxiv.py
```diff
@@ -26,8 +30,22 @@
...
     ]
 
-    # Tar up everything in the src/tex directory
-    with tarfile.open("arxiv.tar.gz", "w:gz") as tarball:
-        for file in paths.user().compile.rglob("*"):
-            if file.name not in exclude and file.suffix != ".bib":
-                tarball.add(file, arcname=file.name)
+    with TemporaryDirectory() as tmpdir:
+        # First, copy to a temporary directory, excluding files
+        shutil.copytree(paths.user().compile, tmpdir, dirs_exist_ok=True)
```
dfm: This copy isn't necessary unless the preprocessing is enabled, I'd move the check above so that we only do this extra copy if needed.

dfm:
If you're not up for adding a test, I think it would be good to at least add this example script to the docs!

showyourwork/workflow/scripts/arxiv.py
MilesCranmer: Fine with me!

showyourwork/workflow/scripts/arxiv.py
MilesCranmer: Sounds good. Will add.

MilesCranmer:
All suggestions implemented! :+1:

MilesCranmer:
Re: a unit-test, I guess I'm just not sure how to approach it in a way that could work in GitHub actions. Seems tricky. But yeah I agree eventually we should add a test for it.

The best test would be a full-scale test of arXiv's compiler on the arXiv tarball, but that would require installing TeXLive in some GitHub action – will be expensive and perhaps not optimal for unittesting (better for some cronjob maybe). I guess in the meantime we have to rely on users reporting any issues with it.

dfm (approved):
Looks good to me! Thanks!

MilesCranmer:
Thanks!! Could a new version (0.4.3) with this change be shipped as well?

PR #365, opened 2023-05-02
Recommend the user use `shell-escape-cwd`

MilesCranmer:
This adds another recommended tectonic arg to the docs. This arg is required to use packages such as `minted` and `biber`.

Otherwise, because `tectonic` builds in a temporary directory, all intermediate outputs and caches will be deleted and therefore not recorded in the arxiv tarball.

With this change and #364 I was finally able to get a working arxiv tarball that uses the `minted` package!

(For any people who find this - the repo https://github.com/MilesCranmer/pysr_paper is a working example of `showyourwork` + `minted`, without needing to do any tweaks to the arxiv tarball.)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1057,12 +1057,14 @@ in the compiled PDF.
 
 .. code-block:: yaml
 
-  tectonic_args: ["-Z", "shell-escape"]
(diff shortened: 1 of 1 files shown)


dfm:
Thanks!

PR #394, opened 2023-09-12
add Schlecker et al. 2023

matiscke:
this adds Schlecker et al. 2023, accepted in PSJ, to the list of papers that use showyourwork.


dfm:
Thanks!

PR #410, opened 2023-10-20
more changes for new Zenodo API

maxisi:
I think this fixes caching!


dfm:
Thanks! We'll see if the tests pass on `main`. There may also be "referrer" issues (#408), but we'll cross that bridge when we get there.

PR #462, opened 2024-07-18
Fix Read The Docs custom domain support

dfm:


HealthyPear:
The reason why this fails on _ReadTheDocs_ might be that we are giving more priority to the `default` conda channel than to `conda-forge`. [In their docs](https://docs.readthedocs.io/en/stable/guides/conda.html#effective-use-of-channels) they recommend to do the opposite or even just use `conda-forge`.

All the packages defined in that recipe are in `conda-forge` and the pinned minimum versions are relatively old, so I would even simplify the environment file as such,

```yaml
channels:
  - conda-forge
dependencies:
  - python
  - datrie
  - requests
  - sphinx-book-theme
  - jinja2
  - sphinxcontrib-programoutput
```

HealthyPear:
@dfm can I try pushing a commit on this branch with the change I suggest above? If it doesn't work we can drop it and resume from your first commit

HealthyPear:
Seems that fixing the conflicts with the main branch was the reason for the failing job

Issue #419, opened 2023-11-30
`\variable` command does not work inside `\caption{...}`


rmnldwg:
First, thanks a lot for this great tool! I really like the idea of making everything except the text itself dynamic.

I am currently writing a paper using showyourwork and have stumbled across something that I couldn't resolve for quite a while: One of my rules kept being ignored and thus it broke my remote builds in particular. Eventually, I figured out that one `\variable{output/...}` command was inside the `\caption{...}` of a figure environment. Apparently, this causes the file inside the `\variable` command not to be part of the workflow graph. Below is a minimal example:

If I have something like this in my main `ms.tex`

```latex
\begin{figure}
    \centering
    \includegraphics[width=\linewidth]{figures/test.png}
    \caption{In this caption, I wanted to include \variable{output/inside.txt}, but it doesn't work.}
    \label{fig:one}
\end{figure}

Outside the figure environment, I can put \variable{output/outside.txt}.
```

and corresponding rules in the `Snakefile`

```
rule inside_figure_env:
    output: "src/tex/output/inside.txt"
    shell: "echo 'some text' > src/tex/output/inside.txt"

rule outside_figure_env:
    output: "src/tex/output/outside.txt"
    shell: "echo 'whatever I want' > src/tex/output/outside.txt"
```

Then the `inside_figure_env` rule will be ignored and the build will fail as long as `src/tex/output/inside.txt` is missing. In my case, I had a Python script that depended on some data. Changing the script or said data also did not trigger a recomputation of the rule as one would expect.

I hope this helps. Thanks a lot for the great work!

arm61:
Can you try wrapping the `\variable` in a `\protect`?

`\variable` is a wrapper round `\input` and I have had some success with this solution before https://tex.stackexchange.com/questions/207905/using-input-inside-caption

rmnldwg:
Unfortunately, that did not work.

I think the issue isn't the `\input` command: If the file `output/inside.txt` in the above example is present, it compiles just fine. The issue is just that the workflow does not get triggered by a `\variable` inside a `\caption`.

What _does_ work is to create an additional dependency in the `showyourwork.yml`:

```yaml
dependencies:
  src/tex/ms.tex:
    - src/tex/output/inside.txt
```

Then, one can also simply use the `\input` command, since adding this dependency implicitly is the main purpose of `\variable`, if I understood that correctly.

arm61:
[...]

PR #485, opened 2024-10-24
Add #419 to known issues

rmnldwg:
As [suggested by HealthyPear](https://github.com/showyourwork/showyourwork/issues/419#issuecomment-2434695335), I added a description of issue #419 to the list of known issues in the docs.

Note that I also added the `hacks` library to the dependencies in the `docs/environment.yml` file, because otherwise `make html` would not have worked.

### DIFF
--- a/docs/known_issues.rst
+++ b/docs/known_issues.rst
@@ -31,3 +31,14 @@ and modify the ``Build the article PDF`` step like this,
       with:
         showyourwork-spec: git+https://github.com/showyourwork/showyourwork
       uses: showyourwork/showyourwork-action@v1
+
+`'\variable' command does not work inside '\caption{...}' (#419) <https://github.com/showyourwork/showyourwork/issues/419>`_
+----------------------------------------------------------------------------------------------------------------------------
+
+When putting the ``\variable{}`` command inside (e.g. a figure's) ``\caption{}``, then the
+corresponding ``Snakefile`` rule will not be triggered.
+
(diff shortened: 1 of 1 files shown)


HealthyPear:
> Note that I also added the hacks library to the dependencies in the docs/environment.yml file, because otherwise make html would not have worked.

This should serve the same purpose of #480 , though I do not understand here: the `hacks` module that `conf.py` imports is not a conda package, but a local file in the `docs` directory (this is why I used `sys`). How does this work here?

rmnldwg:
Oh, I see... Sorry, I didn't even consider that it could be a local file and just assumed `hacks` was a package.

So, how does this work now? Should I add the changes from #480 to this pull request?

HealthyPear:
Technically that change is not needed for this PR as the documentation compiles succesfully online (probably some different setting on readthedocs). #480 is needed for local compilation

HealthyPear:
@rmnldwg this branch fell a bit behind with the default branch

before merging I suggest to cleanup its history by doing this:

- from your branch call `git rebase -i HEAD~3`, this should open your default editor
- change the 1st and 3rd commit from `pick` to `drop`
- save and close the editor
- `git push --force`

then I will rebase this branch on top of the `main` branch and merge it

rmnldwg:
I cleaned up the history as you suggested and rebased it onto the updated `main` branch.

HealthyPear:
Thanks for this!

PR #511, opened 2025-01-14
Add sphinx.configuration key for ReadTheDocs

HealthyPear:
This configuration key is now required by this service.

More info here:
https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,5 +1,8 @@
 version: 2
 
+sphinx:
+  configuration: docs/conf.py
+
 build:
   os: ubuntu-20.04
   tools:


HealthyPear:
@dfm I was sure I already did this, but still the CI check was failing. I understood why: the repo which read the docs is linking is the one under the rodluger user and not the one under the showyourwork organization, so we need to update it but I have not enough writing permissions to do so. Can you do it?

HealthyPear:
@rodluger

HealthyPear:
see #523

PR #518, opened 2025-01-26
Set write permissions to all jobs triggered by action

### LINKED ISSUE #318: Github action default permissions
It looks like now Github by default sets read-only permission on Github actions (unless it is a setting on my account I haven't found).

I see it mentioned in https://show-your.work/en/latest/faqs/#permissions-errors-in-github-actions
however it doesn't only apply to forked repositories.

If this is the case, would it be useful to mention it right in the Quickstart? https://show-your.work/en/latest/quickstart/#build-on-the-remote

HealthyPear:
@dfm could you crosscheck if this is what is needed to make sure that a new project has already the correct permissions?

I still have to test it myself.

If this works, eventually we can update the documentation by removing the [corresponding entry from the FAQ](https://show-your.work/en/latest/faqs/#permissions-errors-in-github-actions).

Closes #318

### DIFF
--- a/src/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/.github/workflows/build-pull-request.yml
+++ b/src/showyourwork/cookiecutter-showyourwork/{{ cookiecutter.repo }}/.github/workflows/build-pull-request.yml
@@ -7,6 +7,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     name: Build the article PDF
+    permissions: write-all
     concurrency: showyourwork-${{ "{{" }} github.ref {{ "}}" }}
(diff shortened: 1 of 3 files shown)


HealthyPear:
OK I have tested this myself by creating two new projects:

the first on using the `main` branch, which fails with the expected message `fatal: unable to access 'https://github.com/HealthyPear/test_showyourwork_main/': The requested URL returned error: 403`

https://github.com/HealthyPear/test_showyourwork_main/actions/runs/12977861085/job/36191760183

while the second one using this PR's branch succeeds

https://github.com/HealthyPear/test_pr_518/actions/runs/12977589168/job/36191171189

PR #555, opened 2025-07-22
Remove main and r channels from integration tests workflow due to new Terms of Service 

HealthyPear:
As #544 shows, the CI is currently failing due to a change upstream related to new Terms of Service for channels we should not even use.

I am following the suggestion from the [logs](https://github.com/showyourwork/showyourwork/actions/runs/16428264132/job/46454068824?pr=554):

> CondaToSNonInteractiveError: Terms of Service have not been accepted for the following channels. Please accept or remove them before proceeding:
>     • https://repo.anaconda.com/pkgs/main
>     • https://repo.anaconda.com/pkgs/r
> 
> To accept a channel's Terms of Service, run the following and replace `CHANNEL` with the channel name/URL:
>     ‣ conda tos accept --override-channels --channel CHANNEL
> 
> To remove channels with rejected Terms of Service, run the following and replace `CHANNEL` with the channel name/URL:
>     ‣ conda config --remove channels CHANNEL

Not sure if the workflow is the place where this is to be changed or if it's a config change for the `snakemake` call of `conda`, but let's see.

### DIFF
--- a/.github/workflows/local_integration_tests.yml
+++ b/.github/workflows/local_integration_tests.yml
@@ -44,7 +44,7 @@ jobs:
         run: |
           conda tos accept --override-channels --channel https://repo.anaconda.com/pkgs/main
           conda tos accept --override-channels --channel https://repo.anaconda.com/pkgs/r
-      
+
       - name: Set channel priority
         shell: bash -l {0}
         run: |
--- a/.github/workflows/local_integration_tests.yml
+++ b/.github/workflows/local_integration_tests.yml
@@ -39,9 +39,18 @@ jobs:
             python=${{ matrix.python }}
             mamba
 
+      - name: Accept Anaconda Terms of Service
+        shell: bash -l {0}
+        run: |
+          conda tos accept --override-channels --channel https://repo.anaconda.com/pkgs/main
+          conda tos accept --override-channels --channel https://repo.anaconda.com/pkgs/r
+      
       - name: Set channel priority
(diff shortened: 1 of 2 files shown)


HealthyPear:
Seems to be the call from _snakemake_, so now I have to find where this is supposed to be set...

HealthyPear:
@dfm is the conda used for the local integration tests the one that is setup from the github worflow?

because I cannot find where is the conda which creates the tectonic.yml env for those test and in there I am already doing

```yaml
    - name: Set channel priority
        shell: bash -l {0}
        run: |
          conda config --remove channels defaults || true
          conda config --remove channels https://repo.anaconda.com/pkgs/main || true
          conda config --remove channels https://repo.anaconda.com/pkgs/r || true
          conda config --add channels conda-forge
          conda config --set channel_priority strict
```

but the error is still there

HealthyPear:
OK removing the channels did not work because (my guess) snakemake takes control of the conda configuration in some way and must re-add them later on

Accepting the ToS instead worked.

In trying to fix this I have introduced some changes that might not be needed, so I will rebase by keeping only what fixes this problem, but there is one thing that I will keep which I discovered running the local test on my laptop (Apple M3 Pro with macos 15.5) which is that the cookicutter default env file for the dummy project install matplotlib via pip and this crashes on my architecture. That environment file is indeed a bit "ancient" by now so I updated it and now the test works both locally and here.

PR #557, opened 2025-07-28
Implement a changelog based on towncrier

### LINKED ISSUE #521: Better changelog
@dfm how are we writing the changelog?

It seems to me that the `HISTORY.rst` file has been edited manually.

I would propose to opt for a more semi-automatic procedure.
I had a nice experience using [towncrier](https://towncrier.readthedocs.io/en/stable/) (and its [sphinx plugin](https://sphinxcontrib-towncrier.readthedocs.io/en/latest/)).

Each entry is generated via a news fragment attached to the PR by its author (the process can be crosschecked via CI to make sure the file is attached to the PR).

The changelog is then automatically updated at each release and you can even show the development version changes before the next release.

In case a PR is related to a change no important to the high level user (such as action bumps) we can use a specific label such as "no-changelog-needed" to disable the CI job and let the PR be merged without a news fragment.

What do you think?

HealthyPear:
This PR adds [towncrier](https://towncrier.readthedocs.io/en/stable/index.html) as a new docs dependency and implements a changelog which is build upon the addition of a "news fragment" to each PR (unless the PR is labelled 'no-changelog-needed').

The changelog will have the sections:

   - ``bugfix``: for bug fixes
   - ``feature``: for new features
   - ``maintenance``: for maintenance tasks, like updating dependencies or documentation
   - ``api``: for changes to the public API
   - ``optmization``: for changes that improve performance or refactor code , but do not change user experience

for each released version of _showyourwork_, but will also show the upcoming changes associated to the development version.

Closes #521.

The changelog page would look something like this:

image


HealthyPear:
@dfm @rodluger I would prefer if we review/merge this PR before #556, so I can merge this there (I do not think I can rebase it) and add a news fragment for that important update in the changelog.

dfm (approved):
Thanks!

Issue #563, opened 2025-08-11
Roadmap to v0.5.0


HealthyPear:
Last release was more or less 2 years ago.
A lot has changed (biggest change e.g. is support for _snakemake 9_), so I think it is time to plan a new minor release.

There are a few small things which I think might be worth fixing before that, so I created a new milestone

https://github.com/showyourwork/showyourwork/milestone/2

If you have more input on what could be added / fixed fro this new release, please let me know here!

HealthyPear:
For anyone passing through here: until this release is published we highly recommend using the development version as there have been several bugfixes and improvements since then!

HealthyPear:
After reviewing the current changelog of the development version 

https://show-your.work/en/latest/changelog/

I think that it is more correct to label the next release as 0.5.0 as this new release will include multiple new features and a change in minimum supported Python version (which might be a breaking change for some projects).

Calling it 0.4.4 would understate both the user-impact and the scope of changes.

HealthyPear:
OK we are very near....

This is the plan I propose:

- I would like to make an `rc` release (aka 'pre-release not ready for production' v0.5.0-rc.1)
- I would like the last PRs to be merged before this release to be #670, #697 , #702

In particular, #697 will fix Snakemake but only with a lower limit (`>=9.18.1`), I would like any release to really **fix** a Snakemake version, and this one of the reasons why I would like to make first an `rc` release (the `dev` version would instead lift any version pinning).

This release will be then out in the wild for a few weeks (?) of test run while we open a new discussion in the Discussion boards for both users and devs to manifest their love (or disgust) for it.

There will be a last PR before v0.5.0 in which I will run `towncrier` to build the official changelog, aka remove the news fragments and update`CHANGES.rst`.
I am not sure if to do it already with the `rc` as I never did it before, not sure what should be best practice in this case: in my current view the `rc` release is just a validation phase, not the final stable release, so I would be tempted to continue aggregating the news fragments by just tagging the `rc` version and wait for 0.5.0 for the final changelog.

Maybe we can use the release UI to make a summary of the big changes and a sort of announcement for the `rc` version?

@vandalt @dfm

HealthyPear:
OK - remember when I said "OK we are very near...."? Well.... 🤦 

Still, the plan is not changed, please take a look at the milestone panel to check what is missing!

There are some testing issues which are not stable and a few bugs to squash, hopefully not too big.
[...]

PR #576, opened 2025-09-15
Update cookiecutter template versioning scheme

### LINKED ISSUE #507: Update the cookiecutter template to fix known issues #435 and #505 automatically
(Premise: I never used coockiecutter)

When calling `showyourwork setup bla/bla` the CLI subcommand runs

```python
# Set up the repo
    cookiecutter(
        str(paths.showyourwork().cookiecutter),
        no_input=True,
        extra_context={
            "user": user,
            "repo": repo,
            "name": name,
            "showyourwork_version": __version__,
            "cache_sandbox_doi": cache_sandbox_doi,
            "overleaf_id": overleaf_id,
            "year": time.localtime().tm_year,
            "action_spec": action_spec,
        },
        overwrite_if_exists=True,
    )
```

then in the template of e.g. the build workflow

```yaml
     - name: Build the article PDF
        id: build
        uses: showyourwork/showyourwork-action@v1
        {%- if cookiecutter.action_spec %}
        with:
          showyourwork-spec: {{ cookiecutter.action_spec }}{% endif %}
        env:
          SANDBOX_TOKEN: ${{ "{{" }} secrets.SANDBOX_TOKEN {{ "}}" }}
          OVERLEAF_TOKEN: ${{ "{{" }} secrets.OVERLEAF_TOKEN {{ "}}" }}
```

which results always in

```yaml
     - name: Build the article PDF
        id: build
        uses: showyourwork/showyourwork-action@v1
        env:
          SANDBOX_TOKEN: ${{ secrets.SANDBOX_TOKEN }}
          OVERLEAF_TOKEN: ${{ secrets.OVERLEAF_TOKEN }}
```

whereas I would expect something like

```yaml
     - name: Build the article PDF
        id: build
       with:
           showyourwork-spec: 0.4.4.dev53+g8e3f5df.d20241022 # or some other dev version
        uses: showyourwork/showyourwork-action@v1
        env:
          SANDBOX_TOKEN: ${{ secrets.SANDBOX_TOKEN }}
          OVERLEAF_TOKEN: ${{ secrets.OVERLEAF_TOKEN }}
```

HealthyPear:
This is an attempt to make template versioning a bit more dynamic and robust.

This is a summary of changes:

- introduced a new flag option to `showyourwork setup` called `--action-version` which specifies the version of the GitHub action (complementary to `--action-spec` which instead specifies the version of showyourwork that the action should use - I would even change the name of the latter in "showyourwork-remote-version" to make it clearer)
- change the default behavior of `--action-spec` to use on the remote the same version of showyourwork that the user is using to create the new project
- set the default behavior of `--action-version` to check if there are new merged PR into the main version (in this case use the latest commit) and if not use the latest release (for now v1)
- add new unit tests to make sure the cookiecutter templates contain valid YAML and have the expected verions.

On top of all local and unit tests passing I also created a [test repository](https://github.com/HealthyPear/test_pr_576) for this PR in place of the remote tests that are currently out of order to check both the build workflows and build PR workflows.

While I wait for reviews I will update the docs and changelog accordingly.

Closes #507.

### DIFF
--- a/src/showyourwork/cli/commands/setup.py
+++ b/src/showyourwork/cli/commands/setup.py
@@ -2,6 +2,7 @@ import shutil
 import time
 from pathlib import Path
 
+import requests
 from cookiecutter.main import cookiecutter
 
 from ... import __version__, exceptions, overleaf, paths
@@ -10,7 +11,7 @@ from ...subproc import get_stdout
 from ...zenodo import Zenodo
 
 
-def setup(slug, cache, overleaf_id, ssh, action_spec):
+def setup(slug, cache, overleaf_id, ssh, action_spec, action_version):
     """Set up a new article repo.
 
     Args:
@@ -20,8 +21,10 @@ def setup(slug, cache, overleaf_id, ssh, action_spec):
         ssh (bool): If True, use SSH to clone the repository. Otherwise, use HTTPS.
         action_spec (str or None): Showyourwork version passed to showyourwork-action in
             `.github/workflows/*.yml`
-
+        action_version (str or None):
+            Version of the showyourwork-action to use in the workflow.
     """
+
     # Parse the slug
     user, repo = slug.split("/")
     if Path(repo).exists():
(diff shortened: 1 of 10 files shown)


arm61:
@HealthyPear I can review this if you would like?

HealthyPear:
> @HealthyPear I can review this if you would like?

Hi! Sure thank you I did not want to bother since I saw you were not active on this repo for a while. I will accept all the help of course!

arm61 (approved):
I have had a good read through the code and everything seems sensible to me. I have *not* built and tested on a test repo, but @HealthyPear's test repo look logical to me.

src/showyourwork/cli/commands/setup.py
```diff
@@ -47,6 +50,59 @@ def callback(code, stdout, stderr):
...
+            pass
+        # Final fallback
+        return "v1"
+
+    if not action_version:
+        action_version = get_latest_action_version()
+
+    # Set action_spec to local showyourwork version if not provided
+    if not action_spec:
+        action_spec = __version__
```
HealthyPear: Actually I see an issue here which might be very wrong

this should not be the version but something like 

`git+https://github.com/showyourwork/showyourwork`

so right now this will make any project remote fail

arm61: Ah, I didn't spot that. How did your test repo pass?

HealthyPear: because since I had to test this specific branch (of the PR) when I created the project I did it passing the spec with the `-a` option as

`showyourwork setup -a git+https://github.com/showyourwork/showyourwork@fix-cookiecutter_template_versions foo/bar`

and so this conditional statement never went through

(which is actually a missing remote test)

before I add the test I need to figure out if we are covering all the cases: the obvious solution is to set 

`git+https://github.com/showyourwork/showyourwork`

am I missing something?

arm61: It makes sense to me... I think

HealthyPear: OK I fixed it, see below in the thread for a detailed explanation

HealthyPear:
OK now it should be correct and way more robust.

Basically the issue was that I was mistakenly using the _version_ of the package to set the action spec in the cookiecutter template, but what you want is something that can be fed to `pip install`.

The basic problem that triggered this PR is that there is no new release since 2+ years (see #563) so many people was installing locally the _development_ version of showyourwork, but the default behavior of the `--action-spec` flag (and thus of the action itself) was to use the latest release on PyPI on the remote which of course made things crash on the remote and people had to manually update their github workflows every time they started a new project.

The behavior now for action spec is to use the version which is locally installed at the time of project creation.

This opened a kind of a pandora box, because the amount of cases is not just release vs dev version!

Also using the version is not the most robust thing to do in the case of a development version because the version file is created at the time of installation and cached after that!

So this PR now uses directly git to asses the status of your local working tree and comparing it to what your branch (any branch!) is tracking on the remote (of any!):

- **Uncommitted changes**: Local working tree has unstaged/staged changes
- **Unpushed commits**: Local branch is ahead of remote tracking branch
- **Both uncommitted and unpushed**: Combination of both scenarios
- **No tracking branch**: Branch exists locally but has no remote tracking branch
- **Development versions**: Versions with .dev, +, .dirty, or rc suffixes

For each case, you will get the "nearest" development version on the remote and a warning will be logged in case some differences have been detected.

For example in my current case my installed version says `0.4.4.dev219+gf6a876582`, meaning that I am 219 commits after v0.4.3 and I ran the installation when I was at commit `f6a876582`.
But the reality is that my git log shows (after rebasing and force-pushing)

```shell
4020b56 (HEAD -> fix-cookiecutter_template_versions, upstream/fix-cookiecutter_template_versions) Add unit tests for action_spec resolution logic
96eaf84 Add news fragment for PR #576
216d433 Add unit tests for workflow templates
b36c151 Update cookiecutter process PR workflow for new versioning scheme
52f50b5 Update cookiecutter build workflow for new versioning scheme
2d1cd76 Update cookiecutter build PR workflow for new versioning scheme
89daea9 Improve version handling in setup command
24afac8 Update .gitignore
```

so not only I am way after that commit, but it does not even exist anymore!

So now if in this case I have pushed all my local commits on this branch, if I do `showyourwork setup foo/bar` the version used by showyourwork (if I do not specify a specific `--action-spec` of course) will be

```yml
uses: showyourwork/showyourwork-action@085c01791d75285222ba2345b93184046f6e9672
        with:
          showyourwork-spec: git+https://github.com/showyourwork/showyourwork@fix-cookiecutter_template_versions
```

meaning use the latest dev version of the action (because it detected commits on the main branch after the latest release) and the latest commit on this very branch because I pushed everything - if this was not the case it would do the same but printing a warning.

😮‍💨

HealthyPear:
Just made a last update:

- the git commands where run in the cwd but of course this is wrong, they should run only where showyourwork is installed!
- in case showyourwork is installed from a release, that would not be a git repo so the command will catch a specific exception (`exceptions.CalledProcessError`) before passing - thereby using the locally installed version - instead of a generic `Exception`

PR #585, opened 2025-09-19
Fix TestOverleaf::test_local

### LINKED ISSUE #578: tests/integration/test_overleaf.py::TestOverleaf::test_local
I found that for the remote test `TestOverleaf.test_local` the solution should be easy: after #467 we did not update our repository secrets, which are still `OVERLEAF_EMAIL` and `OVERLEAF_PASSWORD ` but there should be instead `OVERLEAF_TOKEN` 

@dfm I do not have enough permissions to fix this, because even if I can add a new repositpry secret I do not have the Overleaf account associated to this GitHub oganization (or if we were using @rodluger account)

Here is an example traceback from the remote workflow:

```python
___________________________ TestOverleaf.test_local ____________________________

self = <test_overleaf.TestOverleaf object at 0x7eff2a6db850>

    def test_local(self):
        """
        Test functionality by creating a new repo, customizing it,
        pushing it to GitHub, and awaiting the article build action to
        complete.
    
        """
        try:
            # Disable screen logging info from showyourwork
            self.disable_logging()
    
            # Always run this first
>           self.startup()

tests/integration/helpers/temp_repo.py:279: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/integration/test_overleaf.py:32: in startup
    overleaf.wipe_remote(self.overleaf_id)
../../../micromamba/envs/test-env/lib/python3.11/site-packages/showyourwork/overleaf.py:161: in wipe_remote
    overleaf_token = get_overleaf_credentials()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
```

overleaf_token = 'OVERLEAF_TOKEN', error_if_missing = False

    def get_overleaf_credentials(
        overleaf_token="OVERLEAF_TOKEN",
        error_if_missing=False,
    ):
        """
        Return the user's Overleaf token, stored in env vars.
    
        Args:
            overleaf_token (str, optional): Environment variable containing the
                Overleaf account password. Default ``OVERLEAF_TOKEN``.
            error_if_missing (bool, optional): Default ``False``.
    
        """
        val = os.getenv(overleaf_token, None)
        if val is None or not len(val):
            if error_if_missing:
                level = "error"
            else:
                # This exception is caught in the enclosing scope
                level = "warn"
>           raise exceptions.MissingOverleafCredentials(level=level)
E           showyourwork.exceptions.overleaf.MissingOverleafCredentials

../../../micromamba/envs/test-env/lib/python3.11/site-packages/showyourwork/overleaf.py:79: MissingOverleafCredentials
------------------------------ Captured log call -------------------------------
DEBUG    showyourwork:subproc.py:14 Initialized empty Git repository in /tmp/tmpxau7vvpm/.git/

WARNING  showyourwork:base.py:69 Overleaf credential `OVERLEAF_TOKEN` not found. This should be set as both environment variable and GitHub repository secret.
=============================== warnings summary ===============================
tests/integration/test_overleaf.py::TestOverleaf::test_local
  /home/runner/micromamba/envs/test-env/lib/python3.11/site-packages/showyourwork/exceptions/base.py:69: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
    get_logger().warn(message)

HealthyPear:
This PR fixes this remote test that was not working since quite some time.

This is the summary of changes:

- **IMPORTANT** I guess the Overleaf token associated with the account from showyourwork (or @dfm or @rodluger) expired so this was the main issue; right now this repo and this test point at a blank Overleaf project on my private account so at some point we might want to use the old account again (or if not created, make one for this organization only for testing)
- I did not know how the old test project was made, so I made a **blank**; this triggered an issue where the wipe_remote function did not find files to remove and crashed, so I added a custom callback that gracefully handles the case where there are no files to remove, preventing the error (that was saying "path specifier '*' doesn't match any files").
- that method and pull/push methods were checking for git errors using text strings in english; this is of course not maintainable as e.g. when testing on my machine this was failing because my local is in italian 😛 so I changed the code to use `git status --porcelain` (which outputs machine-readable status)

So, in summary, the test now runs successfully again with **temporary** Overleaf credentials, and the fixes should work for any user regardless of their system's language settings.

I will just rebase/force-push my changes to remove the intermediate commits I used for debugging the remote workflow and then I think we are good to go for a review.

I will not automatically close the underlying issue (#578) when merging - see the thread below why.

### DIFF
--- a/src/showyourwork/overleaf.py
+++ b/src/showyourwork/overleaf.py
@@ -165,43 +165,59 @@ def wipe_remote(project_id, tex=None):
             cwd=cwd,
             secrets=[overleaf_token],
         )
-        get_stdout(["git", "rm", "-r", "*"], cwd=cwd)
+
+        # Try to remove all files, but don't fail if there are no files to remove
+        def rm_callback(code, stdout, stderr):
+            if code != 0:
+                # Check if the error is because no files match the pattern
+                if (
+                    "non corrisponde ad alcun file" in stderr
+                    or "did not match any files" in stderr
+                ):
+                    # This is expected when the repo is empty, so we ignore it
+                    pass
+                else:
+                    # Re-raise other git rm errors
+                    raise exceptions.CalledProcessError(stderr)
+
+        get_stdout(["git", "rm", "-r", "*"], cwd=cwd, callback=rm_callback)
         with open(Path(cwd) / "main.tex", "w") as f:
             print(tex, file=f)
         get_stdout(["git", "add", "main.tex"], cwd=cwd)
 
-        def callback(code, stdout, stderr):
+        # Check if there are changes to commit using git status
+        def status_callback(code, stdout, stderr):
             if code != 0:
-                if (
-                    "Your branch is up to date" in stdout + stderr
-                    or "nothing to commit" in stdout + stderr
-                    or "nothing added to commit" in stdout + stderr
-                ):
-                    pass
-                else:
-                    raise exceptions.CalledProcessError(stdout + "\n" + stderr)
-            else:
-                get_stdout(
-                    ["git", "push", url, "master"],
-                    cwd=cwd,
-                    secrets=[overleaf_token],
-                    callback=check_for_rate_limit,
-                )
+                raise exceptions.CalledProcessError(stderr)
+            return stdout.strip()
 
-        get_stdout(
-            [
-                "git",
-                "-c",
-                "user.name='showyourwork'",
-                "-c",
-                "user.email='showyourwork'",
-                "commit",
-                "-am",
-                "automatic showyourwork update",
-            ],
+        status_output = get_stdout(
+            ["git", "status", "--porcelain"],
             cwd=cwd,
-            callback=callback,
+            callback=status_callback,
         )
+
+        if status_output:
+            # There are changes to commit and push
+            get_stdout(
+                [
+                    "git",
+                    "-c",
+                    "user.name='showyourwork'",
+                    "-c",
+                    "user.email='showyourwork'",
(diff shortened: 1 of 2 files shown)


HealthyPear:
now this repo has my token and I am using a blank project id created by me 

the test still fails with

`2025-09-19T13:19:07.5993692Z stderr = 'remote: {"message":"HTTP error 401"}\nfatal: Authentication failed for '[https://git.overleaf.com/6409f16f438b5fb7c4dfa837/\'\n`

HealthyPear:
OK now the test is passing again! 🎉🎉🎉

HealthyPear:
Damn after I rebased and force-pushed it is failing again with an older project ID, but the diff shows clearly that it should use the new one.

Perhaps it is caused by how the remote workflow is setup...

HealthyPear:
I think the workflow runs the tests from the main branch but using the code from this branch because it fails trying to reach the old overleaf project id.

I think the workflow should check out always the latest commit from the branch were it is run, but this is usually considered a security risk (which is the reason why it has been setup like this I think...)

Since I am not the main maintainer I think I will simply keep the workflow settings as they are and merge to main: the test should pass there and future PRs debugging remote tests will incur in the same issues I faced here...

PR #595, opened 2025-10-20
Add Reep et al paper to list of projects that use `showyourwork`

wtbarnes:
This adds the paper ["Modeling Time-variable Elemental Abundances in Coronal Loop Simulations"](https://iopscience.iop.org/article/10.3847/2041-8213/ad64c3) by @jwreep et al. (including myself) to the list of projects that use `showyourwork`.


HealthyPear:
Thank you!

PR #628, opened 2026-01-16
Add favicon to the docs

vandalt:
Tiny PR I made while geting familiar with the docs: it adds a favicon (tab icon) with the "s" from the showyourwork. I tried a variant with "syw!" from the logo but it felt cramped.


HealthyPear:
Hi @vandalt , thank you for this looks nice! Indeed I can feel that the full logo might be too much

PR #633, opened 2026-01-26
Small updates to the documentation and dependencies

vandalt:
This PR groups a few updates I made while going through the developer docs. Point 2 is a bit of a matter of preference between conda and pip requirements specifications, but I do feel it makes for a more coherent experience between docs and code contributions (see point 3).

1. Datrie is no longer required by snakemake upstream, so I removed the requirement and updated the docs entry to suggest to just update snakemake and showyourwork if people are having issues with datrie.
2. Since datrie is no longer required, it is easy to install all dependencies with pip. I updated the contributor docs to only document the `pip install -U -e ".[docs]"` installation method, which seems simpler and is applicable to "dev" mode and not just "docs" mode.
3. Update the code contribution instruction to mention the creation of an environment and installation of dev dependencies.
4. Add pyyaml as a requirement.

Thanks!

### DIFF
--- a/docs/changes/633.maintenance.rst
+++ b/docs/changes/633.maintenance.rst
@@ -1,0 +1,3 @@
+Removed datrie dependency as it is no longer required upstream.
+Added pyyaml dependency explicitely.
+Updated contributor installation instructions to use pip for docs and for code.
--- a/docs/contributorguide.rst
+++ b/docs/contributorguide.rst
@@ -7,6 +7,8 @@ Contributing to the documentation
 A great place to start contributing is to documentation. In order to do this you'll
 want to run the docs locally. Here are the steps for getting setup.
 
+.. _initial-setup:
+
 Initial setup
 ^^^^^^^^^^^^^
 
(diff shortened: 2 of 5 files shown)


HealthyPear:
> Datrie is no longer required by snakemake upstream, so I removed the requirement and updated the docs entry to suggest to just update snakemake and showyourwork if people are having issues with datrie.

perfect thanks I did not know about thins upstream change

> Since datrie is no longer required, it is easy to install all dependencies with pip. I updated the contributor docs to only document the pip install -U -e ".[docs]" installation method, which seems simpler and is applicable to "dev" mode and not just "docs" mode.

Fine by me. One could even relax the condition and just say something like "any virtual environment with pip" as when you use showyourwork it will pull it's conda anyway

>Update the code contribution instruction to mention the creation of an environment and installation of dev dependencies.

thanks!

> Add pyyaml as a requirement.

Fine, something else was pulling it as a dependency, but better be specific

vandalt:
> One could even relax the condition and just say something like "any virtual environment with pip" as when you use showyourwork it will pull it's conda anyway

Does it pull its own conda or does it use the conda installation that is available on the user's system? My understanding was the latter. In any case I feel like leaving the instruction to create the environment will help users who are not sure of what to do, and those who prefer other methods already know how to create their environment so they can just ignore that part.

HealthyPear:
> Does it pull its own conda or does it use the conda installation that is available on the user's system? My understanding was the latter. 

yes sorry I got confused with the action!

HealthyPear:
Thank you for these updates!

PR #637, opened 2026-01-27
Update `showyourwork` for compatibility with Windows

### LINKED ISSUE #33: Windows compatibility
We should add CI tests on Windows. I'm not sure the `Makefile` approach works on Windows. Also, `Snakemake` syntax for files uses Posix-style paths (even on Windows), but any code that's actually executed within a workflow should use the correct path syntax for the current OS. I've tried to enforce that but it is currently not tested.

### LINKED ISSUE #636: [Bug]: showyourwork fails on Windows
### Preliminary checks

## Problem
Trying to build a paper with `showyourwork` fails on windows. 

## Digging Deeper
The key issue is that current rules in `src\showyourwork\workflow\rules` use the `shell` directive with commands spanning multiple lines. See for example [here](https://github.com/showyourwork/showyourwork/blob/bc0ff456320fd0b5dcb5e8d9df3662cf5284e3f7/src/showyourwork/workflow/rules/preprocess.smk#L59C1-L69C12) and [here](https://github.com/showyourwork/showyourwork/blob/bc0ff456320fd0b5dcb5e8d9df3662cf5284e3f7/src/showyourwork/workflow/rules/compile.smk#L66C1-L76C12). 

While this works on UNIX-based machines (Linux and MacOS), on Windows it runs only the first line after `"""` and the rest is ignored by `snakemake` as shown in [this stackoverflow post](https://stackoverflow.com/a/59004780) and [this issue](https://github.com/snakemake/snakemake/issues/1090)

## Potential Fix
It may not be elegant to have a very long shell command on one line. A cleaner solution is using the `run` directive along with `shell` commands similar to discussion in https://github.com/snakemake/snakemake/issues/1090

### To Reproduce

On a windows machine:

1. Setup a conda environment with `showyourwork` and `git` [(for bash on Windows)](https://snakemake-api.readthedocs.io/en/stable/_modules/snakemake/utils.html#find_bash_on_windows):
  - a. create an `environment.yml` file with the following contents: 
  ```{yml}
  name: showyourwork
  channels:
    - conda-forge
  dependencies:
    - python=3.13
    - pip
    - git
    - pip:
      - git+https://github.com/showyourwork/showyourwork
  ```
  - b. create conda enviroment with `conda env create -f environment.yml`
  - c. activate conda environment with `conda activate showyourwork`

2. Create a dummy `showyourwork` project: `showyourwork setup windows/test`
3. Build the project: 
  - `cd test`
  - `showyourwork build`

 

### Logs

Assuming unrestricted shared filesystem usage.
None
host: WINDOWS-LAPTOP
Building DAG of jobs...
Provided cores: 22
Rules claiming more threads will be scaled down.
Job stats:
job                    count
-------------------  -------
syw__main                  1
syw__preprocess            1
syw__preprocess_xml        1
total                      3
Select jobs to execute...
Execute 1 jobs...
 Rule: syw__preprocess_xml, Jobid: 2
Shell command:
        cd "C:/Users/TEST/Downloads/TEST_PROJECT/.showyourwork/preprocess"
        tectonic                      \
            --chatter minimal         \
            --keep-logs               \
            --keep-intermediates      \
            -r 0                      \
                    \
            "C:/Users/TEST/Downloads/TEST_PROJECT/.showyourwork/preprocess/ms.tex"
Activating conda environment: .snakemake\conda\1dd50042bc273d83900b4eed7a9c7a7f_
Waiting at most 5 seconds for missing files:
C:/Users/TEST/Downloads/TEST_PROJECT/.showyourwork/preprocess/showyourwork.xml (missing locally)

### Setup information

```python
```

```
>>> import platform; print(f"Platform: {platform.platform()}")
Platform: Windows-11-10.0.26200-SP0
>>> import platform; print(f"Python: {platform.python_version()}")
Python: 3.13.11
>>> import showyourwork; print(f"showyourwork: {showyourwork.__version__}")
showyourwork: 0.4.4.dev266+gfc59d738e
```

thealanjason:
Closes #636 
Closes #33


thealanjason:
I can confirm that this works on a Windows machine, but, we might need to add runners for windows in the CI :)

HealthyPear:
The local tests using the window runner are failing, I can see it's because the conda executable cannot be found

are you familiar with windows github runners? maybe iwhen the runner is windows we need to set some setting first or change the path?

HealthyPear:
sorry I mis-clicked 😅

HealthyPear:
I think the problem here is the use of

`shell: bash -l {0}`

I expect windows to not have bash but powershell?

this is the action that is used

https://github.com/mamba-org/setup-micromamba

maybe we could try to set

`shell: micromamba-shell {0}`

and optionally also 

`init-shell: bash powershell`

what do you think?

thealanjason:
> I think the problem here is the use of
> 
> `shell: bash -l {0}`
> 
> I expect windows to not have bash but powershell?
> 
> this is the action that is used
> 
> https://github.com/mamba-org/setup-micromamba
> 
> maybe we could try to set
> 
> `shell: micromamba-shell {0}`
> 
> and optionally also
> 
> `init-shell: bash powershell`
> 
> what do you think?

Thank you for the heads up! 

I also think that this test (that edits the workflow in place) might also be a problem: https://github.com/showyourwork/showyourwork/blob/main/tests/integration/test_latex.py

HealthyPear:
> I also think that this test (that edits the workflow in place) might also be a problem: https://github.com/showyourwork/showyourwork/blob/main/tests/integration/test_latex.py

mmm not sure: before we tried to switch to micromamba shell at least the local tests on the ubunut runner were all succeeding

thealanjason:
In order for window's tests to pass, we need to update them with "..." inside '......' instead of the other way around, just like I did here: https://github.com/showyourwork/showyourwork/pull/637/commits/87fa4ebe3b4504e1c884ff6bd8ec1d03a4509c3e

thealanjason:
> In order for window's tests to pass, we need to update them with "..." inside '......' instead of the other way around, just like I did here: [87fa4eb](https://github.com/showyourwork/showyourwork/commit/87fa4ebe3b4504e1c884ff6bd8ec1d03a4509c3e)

And avoid passing environment variables directly in the command, e.g. "CI=false <command>"

thealanjason:
Things that fail on windows: 
1. Permission issues (for overwriting / deleting files and folders) 
```
 ---------------------------- Captured stdout call -----------------------------
[test-local-cache-optimization] Creating local repo `tests/sandbox/test-local-cache-optimization`...
[test-local-cache-optimization] Building the article locally (1/2)...
---------------------------- Captured stderr call -----------------------------
A subdirectory or file .showyourwork already exists.

Error occurred while processing: .showyourwork.
```
```
            # Delete local repo (only on success)
>           self.delete_local()

tests\integration\helpers\temp_repo.py:301: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests\integration\helpers\temp_repo.py:263: in delete_local
    shutil.rmtree(self.cwd)
C:\Users\runneradmin\micromamba\envs\test-env\Lib\shutil.py:787: in rmtree
    return _rmtree_unsafe(path, onerror)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
C:\Users\runneradmin\micromamba\envs\test-env\Lib\shutil.py:629: in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
C:\Users\runneradmin\micromamba\envs\test-env\Lib\shutil.py:629: in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
C:\Users\runneradmin\micromamba\envs\test-env\Lib\shutil.py:629: in _rmtree_unsafe
    _rmtree_unsafe(fullname, onerror)
C:\Users\runneradmin\micromamba\envs\test-env\Lib\shutil.py:634: in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

path = 'D:\\a\\showyourwork\\showyourwork\\tests\\integration\\sandbox\\test-clean-force\\.git\\objects\\01'
onerror = <function rmtree.<locals>.onerror at 0x000001C584BFBE20>

    def _rmtree_unsafe(path, onerror):
        try:
            with os.scandir(path) as scandir_it:
                entries = list(scandir_it)
        except OSError:
            onerror(os.scandir, path, sys.exc_info())
            entries = []
        for entry in entries:
            fullname = entry.path
            if _rmtree_isdir(entry):
                try:
                    if entry.is_symlink():
                        # This can only happen if someone replaces
                        # a directory with a symlink after the call to
                        # os.scandir or entry.is_dir above.
                        raise OSError("Cannot call rmtree on a symbolic link")
                except OSError:
                    onerror(os.path.islink, fullname, sys.exc_info())
                    continue
                _rmtree_unsafe(fullname, onerror)
            else:
                try:
>                   os.unlink(fullname)
E                   PermissionError: [WinError 5] Access is denied: 'D:\\a\\showyourwork\\showyourwork\\tests\\integration\\sandbox\\test-clean-force\\.git\\objects\\01\\4969f2180258a718372eab62dc495bbfcf3720'

C:\Users\runneradmin\micromamba\envs\test-env\Lib\shutil.py:632: PermissionError
```

2. Dag Rendering fails, because imagemagick is not available on windows (through conda)
```
 E           conda env create --quiet --no-default-packages --file "D:\a\showyourwork\showyourwork\tests\integration\sandbox\test-dag-generation\.snakemake\conda\c423800a340ba1984df9f5fca31582c3_.yaml" --prefix "D:\a\showyourwork\showyourwork\tests\integration\sandbox\test-dag-generation\.snakemake\conda\c423800a340ba1984df9f5fca31582c3_"

E           Output:

E           Channels:

E            - conda-forge

E           Platform: win-64

E           Collecting package metadata (repodata.json): ...working... done

E           Solving environment: ...working... failed

E           

E           PackagesNotFoundError: The following packages are not available from current channels:

E           

E             - imagemagick

```

3. Multiple commands require a space in between: 
```
E           error: halted on potentially-recoverable error as specified
E           RuleException:

E           CalledProcessError in file "D:\a\showyourwork\showyourwork\src\showyourwork\workflow\rules/compile.smk", line 81:

E           Command '"C:/Users/runneradmin/micromamba/envs/test-env/Scripts/activate.bat" "D:/a/showyourwork/showyourwork/tests/integration/sandbox/test-duplicate-graphics/.snakemake/conda/137d12b660efbc0aaaa0c31267c9d519_"&&tectonic --chatter minimal --keep-logs --keep-intermediates --synctex  "D:/a/showyourwork/showyourwork/tests/integration/sandbox/test-duplicate-graphics/.showyourwork/compile/ms.tex"' returned non-zero exit status 1.
```

4. Comparison of Relative v/s Absolute paths: 

```
 tests\integration\helpers\temp_repo.py:298: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests\integration\test_synctex.py:20: in check_build
    path = Path(path).relative_to(self.cwd)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = WindowsPath('/a/showyourwork/showyourwork/tests/integration/sandbox/test-synctex/.showyourwork/compile/ms.tex')
other = (WindowsPath('D:/a/showyourwork/showyourwork/tests/integration/sandbox/test-synctex'),)
parts = ['\\', 'a', 'showyourwork', 'showyourwork', 'tests', 'integration', ...]
drv = '', root = '\\'

    def relative_to(self, *other):
        """Return the relative path to another path identified by the passed
        arguments.  If the operation is not possible (because this is not
        a subpath of the other path), raise ValueError.
        """
        # For the purpose of this method, drive and root are considered
        # separate parts, i.e.:
        #   Path('c:/').relative_to('c:')  gives Path('/')
        #   Path('c:/').relative_to('/')   raise ValueError
        if not other:
            raise TypeError("need at least one argument")
        parts = self._parts
        drv = self._drv
        root = self._root
        if root:
            abs_parts = [drv, root] + parts[1:]
        else:
            abs_parts = parts
        to_drv, to_root, to_parts = self._parse_args(other)
        if to_root:
            to_abs_parts = [to_drv, to_root] + to_parts[1:]
        else:
            to_abs_parts = to_parts
        n = len(to_abs_parts)
        cf = self._flavour.casefold_parts
        if (root or drv) if n == 0 else cf(abs_parts[:n]) != cf(to_abs_parts):
            formatted = self._format_parsed_parts(to_drv, to_root, to_parts)
>           raise ValueError("{!r} is not in the subpath of {!r}"
                    " OR one path is relative and the other is absolute."
                             .format(str(self), str(formatted)))
E           ValueError: '\\a\\showyourwork\\showyourwork\\tests\\integration\\sandbox\\test-synctex\\.showyourwork\\compile\\ms.tex' is not in the subpath of 'D:\\a\\showyourwork\\showyourwork\\tests\\integration\\sandbox\\test-synctex' OR one path is relative and the other is absolute.
```

thealanjason:
The only tests that fail right now are the ones which have a figure with the `\script` latex command. For some reason, on Windows, it fails to be recognized.

thealanjason:
Since imagemagick is not available for windows on conda, I have  switched to using `PyMuPDF` and `pillow` for the pdf->png conversion, to make dag rendering crossplatform.

@HealthyPear what are your thoughts?
 Also, would you want me to split this into a separate issue?

thealanjason:
> The only tests that fail right now are the ones which have a figure with the `\script` latex command. For some reason, on Windows, it fails to be recognized.

I suspect something in the `add_figure_environment` to be incompatible with windows: https://github.com/showyourwork/showyourwork/blob/bc0ff456320fd0b5dcb5e8d9df3662cf5284e3f7/tests/integration/helpers/temp_repo.py#L371C1-L383C10

thealanjason:
Today I found some time to test the quickstart guide on a windows machine, and I get the following error: 
```
error: ms.tex:67: Undefined control sequence
! Undefined control sequence.
\syw@data@fig:random_numbers_script ->src\scripts
                                                  \random _numbers.py
l.67         \label{fig:random_numbers}

No pages of output.
Transcript written on ms.log.
error: halted on potentially-recoverable error as specified
```

HealthyPear:
> Today I found some time to test the quickstart guide on a windows machine, and I get the following error:
> 
> ```
> error: ms.tex:67: Undefined control sequence
> ! Undefined control sequence.
> \syw@data@fig:random_numbers_script ->src\scripts
>                                                   \random _numbers.py
> l.67         \label{fig:random_numbers}
> 
> No pages of output.
> Transcript written on ms.log.
> error: halted on potentially-recoverable error as specified
> ```

I think this might be related to the same issue you experience in add_figure_environment: not an expert of windows, but if I recall correctly paths use the opposite backslash? so that when latex reads the path finds pieces of a string that it interprets like a command (that of course does not exist)

> Since imagemagick is not available for windows on conda, I have switched to using PyMuPDF and pillow for the pdf->png conversion, to make dag rendering crossplatform.
@HealthyPear what are your thoughts?
Also, would you want me to split this into a separate issue?

are you installing PyMuPDF and pillow via conda as we were doing for imagemagick? I have no preference for either of them as long as the associated tests pass and we keep the current approach

thealanjason:
> I think this might be related to the same issue you experience in add_figure_environment: not an expert of windows, but if I recall correctly paths use the opposite backslash? so that when latex reads the path finds pieces of a string that it interprets like a command (that of course does not exist)

Yes, I suspect the same. Tried a fix using \detokenize, waiting for tests to pass (windows tests are really slow)

thealanjason:
> are you installing PyMuPDF and pillow via conda as we were doing for imagemagick? I have no preference for either of them as long as the associated tests pass and we keep the current approach

Yes, I currently install it via conda. However, PyMuPDF is installed via pip on conda. I think this should be ok. I feel that rendering a dag is not really a core-feature of `showyourwork`, but more of an add-on.

thealanjason:
This PR is now ready to be reviewed. @HealthyPear @vandalt could you have a look?

thealanjason:
Ping @HealthyPear :)

HealthyPear:
> Ping @HealthyPear :)

yep I'll test it asap (I do not have a windows machine, but I can test the quickstart example using your branch)

thealanjason:
> > Ping @HealthyPear :)
> 
> yep I'll test it asap (I do not have a windows machine, but I can test the quickstart example using your branch)

Thanks. I went on to make the tests compatible with Windows, and it now runs on CI. So if a test is passing in the CI for both linux and windows, we can be confident that the behaviour is the same.

HealthyPear:
Hi @thealanjason ,

I tested your branch on my macos and the whole quickstart went through as expected, DAG generation included.

Thank you for all this work!

If you agree we can merge this PR to main (I would use a merge commit in place of a rebase as there are many commits with ambiguous messages, with revert or debug purposes, if you agree).

thealanjason:
> If you agree we can merge this PR to main (I would use a merge commit in place of a rebase as there are many commits with ambiguous messages, with revert or debug purposes, if you agree).

Sure we can squash and merge into main :)

HealthyPear:
Thank you!!!

PR #635, opened 2026-01-26
Fix Zenodo integration

vandalt:
See #599 for a description of the issue.

The fix I'm working on right now is to patch the `store` and `fetch` methods on the class definition directly instead of on the instance, since the instance is created only after. Now I'm encountering auth issues with Zenodo that I will need to investigate, so I'm making this a draft.

Fixes #605 
Fixes #440
Fixes #439 
Fixes #386 
Fixes #340

PR #641, opened 2026-01-31
Attempt to fix `TestMatplotlibLatex` remote

vandalt:
In addition to zenodo remote tests in #635, `TestMatplotlibLatex` was failing due to a missing underscore.sty file. I think installing the package should fix the issue, but we would need to run the remote tests on this PR to know for sure (since running locally uses my own tex installation).

Thanks!

### DIFF
--- a/docs/changes/641.bugfix.rst
+++ b/docs/changes/641.bugfix.rst
@@ -1,0 +1,1 @@
+Install underscore.sty latex package in test_latex.py
--- a/docs/changes/641.bugfix.rst
+++ b/docs/changes/641.bugfix.rst
@@ -1,1 +1,0 @@
-Install underscore.sty latex package in test_latex.py
--- a/tests/integration/test_latex.py
+++ b/tests/integration/test_latex.py
@@ -13,7 +13,7 @@ workflow_replace_with = """      - name: Install TinyTex for matplotlib LaTeX re
         shell: bash -l {0}
         run: |
           wget -qO- "https://yihui.org/tinytex/install-bin-unix.sh" | sh
-          sudo ~/bin/tlmgr install type1cm cm-super
+          sudo ~/bin/tlmgr install type1cm cm-super underscore
 
       - name: Build the article PDF
 """


vandalt:
@HealthyPear is it possible to prevent the bot from removing `safe to test` so I can iterate on the tests?

HealthyPear:
> @HealthyPear is it possible to prevent the bot from removing `safe to test` so I can iterate on the tests?

the removal is needed because the tests are triggered by the addition of the label, if it stays there you cannot launch them again

docs/changes/641.bugfix.rst
```diff
@@ -0,0 +1 @@
+Install underscore.sty latex package in test_latex.py
```
HealthyPear: no need for changelog in this case, it's an internal thing

HealthyPear (changes requested):

HealthyPear:
the specific error seems to be

```
2026-02-02T08:59:50.6204633Z RuntimeError: latex was not able to process the following string:
2026-02-02T08:59:50.6204978Z b'lp'
```

this is the actual worklow log from the remote repo

https://productionresultssa16.blob.core.windows.net/actions-results/32bce3bd-9a2c-4154-87b7-7513b76d5cb1/workflow-job-run-d6ecf576-b6c0-58db-9a11-5c3c0d3f2f70/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-02-02T09%3A21%3A32Z&sig=fg1bBC2GuTpHUR7v3xtSc7BXxqxzlAveYuWiAWOn%2BgU%3D&ske=2026-02-02T10%3A36%3A42Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-02-02T06%3A36%3A42Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-02-02T09%3A11%3A27Z&sv=2025-11-05

The local test passes, though, I wonder what is the difference...

HealthyPear:
Ah I still also see

`2026-02-02T08:59:50.6213188Z ! LaTeX Error: File `underscore.sty' not found.`

I am not sure it is worth adding it via the package manager.

Since we use tectonic, it should be sufficient to add directly the \usepackage directive and tectonic should fetch it, but again no reason for the local test to pass no?

HealthyPear:
Aha. I found the underlying issue

```
2026-02-02T08:58:05.0072252Z From https://github.com/showyourwork/test-matplotlib-latex
2026-02-02T08:58:05.0073377Z  * [new branch]      main       -> origin/main
2026-02-02T08:58:05.0074041Z  * [new branch]      main-pdf   -> origin/main-pdf
2026-02-02T08:58:05.0116749Z [command]/usr/bin/git branch --list --remote origin/main
2026-02-02T08:58:05.0140031Z   origin/main
2026-02-02T08:58:05.0150477Z [command]/usr/bin/git rev-parse refs/remotes/origin/main
2026-02-02T08:58:05.0170909Z ab79fba1d7683d3cca5954b6c2e154f123367534       <------------ !!!!!!!
```

this is not your last commit in this PR which is `766c2692c45a9c9876eb85d6c1965545e67dcd4a`

vandalt:
> the specific error seems to be
> 
> 2026-02-02T08:59:50.6204633Z RuntimeError: latex was not able to process the following string:
> 2026-02-02T08:59:50.6204978Z b'lp'

I also saw this one but my understanding is it's a generic error when something goes wrong and that the underlying tex issue is the problem. I might be wrong though.

> The local test passes, though, I wonder what is the difference...

Regarding the local test passing, `build_local()` is overwritten by the class to be a no-op so the local test passes automatically

> Ah I still also see
> 
> 2026-02-02T08:59:50.6213188Z ! LaTeX Error: File underscore.sty' not found.`
> 
> I am not sure it is worth adding it via the package manager.
> 
> Since we use tectonic, it should be sufficient to add directly the \usepackage directive and tectonic should fetch it, but again no reason for the local test to pass no?

My understanding is showyourwork uses tectonic to build the article, but it will not work directly with matplotlib, so we need tinytex to install latex and a few packages. See [this FAQ entry in the docs](https://show-your.work/en/latest/faqs/#rendering-latex-in-matplotlib).

>     2026-02-02T08:58:05.0170909Z ab79fba1d7683d3cca5954b6c2e154f123367534       <------------ !!!!!!!
> this is not your last commit in this PR which is 766c2692c45a9c9876eb85d6c1965545e67dcd4a

The commit used is from the [test-matplotlib-latex](https://github.com/showyourwork/test-matplotlib-latex) repo which gets created during the test (I think). The build file still seems to point to the correct location ([link](https://github.com/showyourwork/test-matplotlib-latex/commit/ab79fba1d7683d3cca5954b6c2e154f123367534#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R18)) which was the latest commit on this branch.

I'm not quite sure how to fix this. One could naively hope that merging this into main would add `underscore` to the installed packages and fix everything, but I don't know if that's true (and actually understanding what happens would be better than just hoping this fixes things).

HealthyPear:
What is explained in the FAQ is indeed the same as in this test (apart from the missing underscore package) so apparently this should already work...

if I click on the link in the build file you posted I do not see your last version though, the underscore package in the tlmgr command is effectively missing

I think the problem is that the remote integration tests workflow pulls only from main: take a look at the beginning

https://github.com/showyourwork/showyourwork/actions/runs/22315565679/job/64559343177?pr=641

under Checkout / Fetching the repository

so it could be indeed that we have to merge to main first...

vandalt:
If you are OK with merging to see if this fixes the issue I think that would be the simplest path forward @HealthyPear.

The PR only contains a tiny change to a test file so it is low risk and would be easy to overwrite. If it does not fix things, I can keep debugging in a new PR.

Thanks!

HealthyPear (approved):

Issue #645, opened 2026-02-24
TestMatplotlibLatex::test_remote is failing


HealthyPear:
OK after #641 was merged now I can see that we are effectively doing the right thing, namely

`sudo ~/bin/tlmgr install type1cm cm-super underscore`

though the workflow is failing due to a new error:

```shell
tlmgr: setting option sys_bin to /home/runner/bin.
tlmgr: updating /home/runner/.TinyTeX/tlpkg/texlive.tlpdb

/home/runner/bin/tlmgr: TLPDB::from_file could not get texlive.tlpdb from: https://ctan.math.illinois.edu/systems/texlive/tlnet/tlpkg/texlive.tlpdb
Maybe the repository setting should be changed.
More info: https://tug.org/texlive/acquire.html
Error: Process completed with exit code 1.
```

See

https://github.com/showyourwork/test-matplotlib-latex/actions/runs/22357820112/job/64702363912

for details

HealthyPear:
@vandalt

HealthyPear:
seems to be a mirroring issue

we should maybe specify a safe mirror explicitely (and maybe force an update)

and sudo should not be needed as we install in the runner's home

see https://tug.org/texlive/doc/tlmgr.html#option

I suggest to try with

```yaml
- name: Install TinyTeX
  run: |
    wget -qO- "https://yihui.org/tinytex/install-bin-unix.sh" | sh
    ~/bin/tlmgr option repository https://mirror.ctan.org/systems/texlive/tlnet
    ~/bin/tlmgr update --self
    ~/bin/tlmgr install type1cm cm-super underscore
```

HealthyPear:
This is now passing on main, I guess it was just a hiccup! IF it happens again we have a possibly stricter solution ready

PR #647, opened 2026-02-24
Update TinyTex setup in TestMatplotlibLatex

HealthyPear:
Update tlmgr commands to set repository and update before installing packages.

### DIFF
--- a/tests/integration/test_latex.py
+++ b/tests/integration/test_latex.py
@@ -13,7 +13,9 @@ workflow_replace_with = """      - name: Install TinyTex for matplotlib LaTeX re
         shell: bash -l {0}
         run: |
           wget -qO- "https://yihui.org/tinytex/install-bin-unix.sh" | sh
-          sudo ~/bin/tlmgr install type1cm cm-super underscore
+          ~/bin/tlmgr option repository https://mirror.ctan.org/systems/texlive/tlnet
+          ~/bin/tlmgr update --self
+          ~/bin/tlmgr install type1cm cm-super underscore
 
       - name: Build the article PDF
 """


HealthyPear:
I launched the pipeline anyway just to see, but I think only `main` matters in this case...

vandalt:
I guess merging is again the only way to see if this works? And #645 can be used to track the remote test runs that you launch manually following merge.

HealthyPear:
Whoa!

`tests/integration/test_latex.py::TestMatplotlibLatex::test_remote PASSED [ 57%][test-synctex] Workflow completed successfully.`

https://github.com/showyourwork/showyourwork/actions/runs/22359512570/job/64708603974?pr=647

HealthyPear:
I am very confused now.

It seems the remote tests installed showyourwork from commit

b26b474ed49aa5926e87d260411bd4bff0507c20

which is this PR

but the workflow file of the last (succesfull) build of the test repo created by this test shows the old version of the tinytex installation

https://github.com/showyourwork/test-matplotlib-latex/actions/runs/22359660798/workflow

but installing showyourowrk from this PR!

with:
          showyourwork-spec: git+https://github.com/showyourwork/showyourwork.git@b26b474ed49aa5926e87d260411bd4bff0507c20#egg=showyourwork

and then it works without the error shown in #645

HealthyPear:
to make sure things are already (magically) working I will manually trigger a remote test workflow on `main` and see if this test passes

perhaps it was just a hiccup?

should we merge this to make extra sure it will not happen again?

HealthyPear:
seems I cannot trigger it manually...

HealthyPear:
I'll wait to see if the test passes on main after #635 - in case it odes I guess it was just a hiccup and we can close this

HealthyPear:
This is now passing on main, I guess it was just a hiccup! IF it happens again we have a possibly stricter solution ready

PR #649, opened 2026-02-25
Fix `showyourwork build --dry-run`

### LINKED ISSUE #648: [Bug]: Snakemake's `--dry-run` argument causes showyourwork to crash
### Preliminary checks

I would expect `--dry-run` to just list the steps as would normally happen for a snakemake workflow.

Currently `showyourwork build --dry-run` fails with an a `MissingConfigFile` error:

```python
An error occurred while executing the workflow.
MissingConfigFile in file "/home/vandal/repos/astro/showyourwork/src/showyourwork/workflow/build.smk", line 114.
  File "/home/vandal/repos/astro/showyourwork/src/showyourwork/workflow/build.smk", line 114, in <module>
```

The issue is that if the `config.json` is not generated by prep.smk, then the error gets raised when the run type is not `clean`. I have a small fix ready where we also check `not workflow.output_settings.dryrun`.

I'll open a PR but wanted to report here first.

### To Reproduce

1. Go to any showyourwork repository
2. Run `showyourwork build --dry-run`

### Logs

_No response_

### Setup information

```
Platform: Linux-6.18.12-200.fc43.x86_64-x86_64-with-glibc2.42
Python: 3.14.2
showyourwork: 0.4.4.dev278+g493b74a18
```

vandalt:
Fixes #648.

This simply checks that we are not in a `--dry-run` before raising the `MissingConfigFile` error.

It will not, however, make the `--dry-run` much more informative: since `config.json` does not get generated basically nothing happens and a dry-run will always say there is nothing to do. This could be confusing so I would either:

1. Fix the dry-run so that a (potentially simpler) version of `prep.smk` always runs even during dry runs, so config.json gets created. This would let snakemake handle the dry-run as usual.
2. At least show a warning saying that currently dry-runs are not representative of what would actually happen.

### DIFF
--- a/docs/changes/649.bugfix.rst
+++ b/docs/changes/649.bugfix.rst
@@ -1,0 +1,1 @@
+Do not throw MissingConfigFile error when `--dry-run` is passed to Snakemake.
--- a/docs/known_issues.rst
+++ b/docs/known_issues.rst
@@ -88,3 +88,10 @@ Please comment line 405 in
 
 Once the Zenodo cache is frozen, no new Zenodo drafts are created by |showyourwork|.
 As a workaround while we work on a solution, we recommend users do not freeze the cache until it is ready to be published.
+
+`Snakemake's --dry-run argument completely skips the workflow #653 <https://github.com/showyourwork/showyourwork/issues/653>`_
+-----------------------------------------------------------------------------------------------------------
+
+showyourwork passes the `--dry-run` argument to snakemake, but internally it will skip the entire workflow.
+This results in a message saying no steps need to be run even when, in fact, steps would be run in an actual build.
+There is a warning printed at the CLI about this.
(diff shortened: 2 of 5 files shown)


vandalt:
I implemented 2 since it's much simpler. I'm looking into 1 right now, but happy to do it in a new PR if you want to merge this sooner @HealthyPear.

HealthyPear:
I did not even know there was this option, but I guess it's because it is actually an option that is fed to snakemake right?

I guess this missing info is only a problem on the showyourwork side. For the moment I am fine with option 2, the priority should be in fixing the crash first and foremost because that is what the user will encounter, then it is only fair we issue at least the working to address the inevitable confusion as you correctly stated

vandalt:
I just pushed a fix for a failing test and added an entry to the changelog. I'll also add the "nothing happens" issue to known issues and then I think it will be ready for merge.

HealthyPear:
@vandalt do you need to add something else to this PR? LFTM

vandalt:
Yes @HealthyPear, thanks!

Issue #459, opened 2024-05-24
Alternative way to run remote tests


HealthyPear:
While working on some features I wanted to make sure that I wasn't breaking anything and when launching the tests I noticed that many of them need to work from a remote repository stored under this organization.

I do not have writing permissions, so I cannot run those tests.

Though I do not think you need to push test repos in your organization: I propose to add an alternative approach to run the remote tests by running the GitHub Action workflows locally.

A solution I used in the past is e.g. https://nektosact.com/ .

HealthyPear:
I think #654 has provided a reasonable solution to this, given it ensures complete control to the user and does not require another third-party software. Closing!

PR #654, opened 2026-02-26
Add option to run remote tests outside of showyourwork org

vandalt:
This PR adds a command line option to run remote tests with another GitHub org or as a regular user. The goal is to make it easier for contributors to run and debug the remote tests without having to rely on marking PRs safe to test.

For example, one could do this to run the default `test_remote` as a GitHub user.

```bash
python -m pytest tests/integration/test_default.py --remote --action-spec git+https://github.com/showyourwork/showyourwork.git -k test_remote --no-org
```

This requires users to setup an API token. I'll update the docs to mention this before marking the PR as ready for review.

### DIFF
--- a/docs/changes/654.maintenance.rst
+++ b/docs/changes/654.maintenance.rst
@@ -1,0 +1,2 @@
+Add command line arguments to run the remote test as a regular user (`--no-org`) or with a specific org (`--github-org`).
+This allows users and contributors to run the remote tests without having to push a PR, making debugging easier.
--- a/docs/changes/654.maintenance.rst
+++ b/docs/changes/654.maintenance.rst
@@ -1,2 +1,2 @@
-Add command line arguments to run the remote test as a regular user (`--no-org`) or with a specific org (`--github-org`).
+Add command line arguments to run the remote tests as a regular user (`--no-org`) or with a specific org (`--github-org`).
 This allows users and contributors to run the remote tests without having to push a PR, making debugging easier.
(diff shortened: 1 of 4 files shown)


HealthyPear:
I also explored alternative ways to do this, see #459 

but I think your solution is more in line with what we actually want to test!

I think it might be worth in this case to add a news fragment under the `maintenance` label so show potential contributors that this is now possible

so, just to make sure you understand, with this option your are still pushing to a remote repo but under your user?

vandalt:
> I think it might be worth in this case to add a news fragment under the maintenance label so show potential contributors that this is now possible

Done!

> so, just to make sure you understand, with this option your are still pushing to a remote repo but under your user?

Yes! See [here](https://github.com/vandalt/test-cache/actions/runs/22429077743) for an example, which I used for #655.

docs/changes/654.maintenance.rst
```diff
@@ -0,0 +1,2 @@
+Add command line arguments to run the remote test as a regular user (`--no-org`) or with a specific org (`--github-org`).
```
HealthyPear: to run any remote test

HealthyPear (changes requested):

docs/changes/654.maintenance.rst
vandalt: Oops, I meant this do be plural (the remote tests). Fixed! @HealthyPear

HealthyPear:
mmm for some reason the readthedocs build succeeded but the communication to github is not arriving

vandalt:
Weird indeed. But I'd be surprised if the changes I made in the PR are the cause here though.

HealthyPear (approved):

PR #657, opened 2026-02-26
Properly pass `--dry-run` to snakemake for build step

### LINKED ISSUE #653: [Bug]: Snakemake's `--dry-run` argument completely skips the workflow
### Preliminary checks

Following #649, `--dry-run` no long throws an error, but it completely skips the workflow and always says there is nothing to do.

This happens because `prep.smk` does not get executed so `config.json` does not exist in `build.smk`.

The expected behaviour would be that all steps that need to be run are printed to the user.

### To Reproduce

1. Install showyourwork
2. Run `showyourwork build --dry-run` in any showyourwork project.

vandalt:
This fixes #653 by passing the `--dry-run` flag to snakemake for the build step. Here is what happens:

1. `--dry-run` or `-n` gets filtered out for the `prep.smk` step: we need this step to run to generate `config.json`.
2. `--dry-run` gets passed to `snakemake` like any other argument for the `build.smk` step so that the workflow runs in dry-run mode as any other snakemake workflow would.

By default, `showyourwork` suppresses Snakemake outputs, so dry-runs will not be very informative. To see more info, users can:

- Look at `.showyourwork/logs/snakemake.log`
- Set `verbose: true` in `showyourwork.yml`
- Pass `--verbose` as a CLI argument.

Before this PR, passing `--verbose` would only affect the output of the `snakemake.log` file unless `verbose: true` was set in the YAML file. This seems counter-intuitive from a user perspective: if I pass `--verbose`, I usually expect the program to print more things to STDOUT. I changed the code slightly so that passing `--verbose` to snakemake would also print snakemake output to STDOUT. This also allows users to quickly check the snakemake output for a single build execution without messing with their config.

### DIFF
--- a/docs/changes/649.bugfix.rst
+++ b/docs/changes/649.bugfix.rst
@@ -1,1 +1,1 @@
-Do not throw MissingConfigFile error when `--dry-run` is passed to Snakemake.
+Do not throw MissingConfigFile error when ``--dry-run`` is passed to Snakemake.
--- a/docs/changes/657.feature.rst
+++ b/docs/changes/657.feature.rst
@@ -1,0 +1,2 @@
+Execute ``prep.smk`` even with ``--dry-run`` and properly pass ``--dry-run`` to snakemake for the build step.
+Ensure that the snakemake output gets printed to STDOUT when ``--verbose`` is passed at the CLI.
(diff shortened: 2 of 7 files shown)


HealthyPear (approved):
LFTM! Thank you!

PR #658, opened 2026-02-27
Raise an error if config dependencies are missing or mistyped

### LINKED ISSUE #242: Warn if file in showyourwork.yml/dependencies is missing
I spent a while trying to figure out why my dependency graph wasn't working like I expected. I eventually realized that in `showyourwork.yml/dependencies`, the name of the script had a typo, and so showyourwork wasn't picking up on the dependencies for that script. It might be helpful if showyourwork checked that all of the files you have listed dependencies for exist, and if they don't, warn you about that?

### LINKED ISSUE #620: Warn if \script calls a script with a mispelled name wrt existing scripts and/or Snakefile scripts
> @dfm I did put a simple version together, and that worked. syw is working as intended.
> 
> 
> 
> I finally found it was my fault - the filename of the script in the `\script{}` latex command was partially capitalized, but the version in the `Snakemake` file was not.  

 _Originally posted by @mkenworthy in [#421](https://github.com/showyourwork/showyourwork/issues/421#issuecomment-1846709714)_

Since I stumbled upon this error myself multiple times, I think it would be a good idea to have a layer in _showyourwork_ where we warn the user about this and (optionally) a `treat_warnings_as_errors` mode.

There are some algorithms one can use to check for misspelling in strings we could look up.

HealthyPear:
I added a new function to check the contents of the `dependencies` section from `showyourwork.yml` in order to check first if there are missing scripts or mistyped names (the core of issue #242).

I also added a similar function to the `dag` workflow because in there the entire workflow is generated, so we can also check if there are entries (values) that do not correspond to something that should be there from the start (e.g. under `src/static`) or stuff that should be generated by the user (via a rule in the `Snakefile`).

It's a delicate PR, so I will wait for someone else to test it further (@jfcrenshaw your issue was very old, but I ping you anyway just in case you still need it).

Closes #242
Closes #620


HealthyPear:
@thealanjason the only tests failing here are related to the windows runner

I'll take a look asap, but I suspect it has something to do with the lookup of files in the file system and management of paths

if you find the issue before me, feel free to post it here or test this branch locally an push directly if you are sure!

tests/integration/test_dependency_warnings.py
```diff
@@ -0,0 +1,151 @@
...
+        self.add_figure_script()
+        self.add_figure_environment()
+
+        # Add a dependency that refers to a non-existent generated file
+        config_path = self.cwd / "showyourwork.yml"
+        with open(config_path) as f:
+            config = yaml.safe_load(f)
+
+        config["dependencies"] = {
+            "src/scripts/test_figure.py": [
```
thealanjason, line 80: Looking closely at the CI output, this test fails because it does not find the `src/scripts/test_figure.py`, and already throws the error on line https://github.com/showyourwork/showyourwork/pull/658/changes#diff-e1632271700bb41a77b7bb6ac1c86c603063f1207f94e889e8ec2b7a69b8f6b4R523

It does not get a chance to proceed with checking whether the actual data is present or not. 

I sense that this is a path separator issue.

src/showyourwork/workflow/scripts/preprocess.py
```diff
@@ -487,6 +488,51 @@ def get_json_tree(xmlfile):
...
+    known_scripts = set()
+    for fig_info in tree["figures"].values():
+        if fig_info["script"] is not None:
+            known_scripts.add(fig_info["script"])
+
+    # The manuscript TeX file is also a valid dependency key
+    known_scripts.add(ms_tex)
+
+    for key in dependencies:
+        if key not in known_scripts:
```
thealanjason: maybe we can use pathlib to generate the paths of known scripts, and then use it to match the paths instead of just matching the path string with `/` as the path separator

thealanjason (changes requested):
Could we use `pathlib` to create and check equal paths?

src/showyourwork/workflow/scripts/preprocess.py
HealthyPear: yep that seemed to be the issue

I now normalized all paths to POSIX paths

PR #665, opened 2026-03-02
Update TestPullRequests

### LINKED ISSUE #650: tests/integration/test_pr.py::TestPullRequests::test_remote
The workflow finds the bot comment and extracts `diff_url` via regex

Later it fails with

`Exception: Diff PDF was not pushed to the -pdf branch.`

In the same test output, there’s also evidence of intermittent GitHub API trouble:

`r = <Response [502]>` shown in the traceback area (coming from `parse_request`), which tells me this might be a GitHub API/hosting issue.

Even when the workflow completed successfully, the PDF hosted from the `-pdf` branch can lag for seconds or minutes due to branch push propagation and caching.

I think the problem here is given by the fact that we are trying to retrieve the `diff-pdf` with a single `request.get` call

```
response = requests.get(diff_url)
        if response.status_code > 204:
            raise Exception("Diff PDF was not pushed to the -pdf branch.")
        with open(self.cwd / "diff.pdf", "wb") as f:
            f.write(response.content)
```

Since the test already waits/polls for workflow completion maybe it should also wait/poll for the hosted artifact to become available?

### LINKED ISSUE #668: [Bug]: diff PDF not generated during Pull Requests
### Preliminary checks

Any PR will allow you to see the PDF with highlighted changes (diff PDF)

see https://show-your.work/en/latest/action/#pull-requests

as far as I can see the diff PDF is not even generated

The respective remote test is currently failing (#650)

### To Reproduce

Given a project, make a change in a new branch and push it to make a PR

you should get 2 workflows: 

- build
- build pull request

a 3rd worfklow should process the PR and generate a message via a bot

this bot will say something like

> Here is the compiled [article PDF](https://github.com/HealthyPear/test_showyourwork_pr/raw/pull-request-1-pdf/ms.pdf), as well as the [PDF with highlighted changes](https://github.com/HealthyPear/test_showyourwork_pr/raw/pull-request-1-pdf/diff.pdf), for the latest commit. You can find additional build output at [pull-request-1-pdf](https://github.com/HealthyPear/test_showyourwork_pr/tree/pull-request-1-pdf).

if you click on the diff PDF link you will get a 404 Page not Found error

### Logs

_No response_

### Setup information

Platform: macOS-15.7.4-arm64-arm-64bit-Mach-O
Python: 3.13.12
showyourwork: 0.4.4.dev316+gfc97b8c18

HealthyPear:
-  poll the artifact diff URL for a short period before failing, instead of a single GET
- use pytest.fail with more information in case of failure

I closed #664 and renamed the branch from `fix-TestPullRequests` to `staging-staging-fix-TestPullRequests` which should allow the remote tests to be triggered by changes of this new branch without waiting to merge to main (@vandalt for future ref)

Closes #650 
Closes #668

### DIFF
--- a/docs/changes/665.bugfix.rst
+++ b/docs/changes/665.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed a bug that prevented a Pull Request to produe the PDF file with
+highlighted changes (see `this issue <https://github.com/showyourwork/showyourwork-action/issues/24>`_ for the original cause).
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -23,8 +23,14 @@ def pytest_addoption(parser):
     parser.addoption(
         "--action-spec",
         action="store",
-        default="showyourwork",
+        default="",
         help="version spec of showyourwork to install on GH Actions",
+    )
+    parser.addoption(
+        "--action-version",
+        action="store",
+        default="",
+        help="version of the showyourwork-action to use in the workflow",
     )
     parser.addoption(
         "--github-org",
@@ -43,6 +49,7 @@ def pytest_addoption(parser):
 def pytest_configure(config):
     config.addinivalue_line("markers", "remote: a test that requires remote access")
     os.environ["ACTION_SPEC"] = str(config.getoption("--action-spec"))
+    os.environ["ACTION_VERSION"] = str(config.getoption("--action-version"))
 
     # Set GITHUB_ORG from CLI option or environment variable
     if config.getoption("--no-org"):
--- a/tests/integration/helpers/__init__.py
+++ b/tests/integration/helpers/__init__.py
@@ -1,4 +1,6 @@
 from .temp_repo import (
+    GITHUB_ORG as GITHUB_ORG,
+    GITHUB_USER as GITHUB_USER,
     ShowyourworkRepositoryActions as ShowyourworkRepositoryActions,
     TemporaryShowyourworkRepository as TemporaryShowyourworkRepository,
 )
--- a/tests/integration/helpers/temp_repo.py
+++ b/tests/integration/helpers/temp_repo.py
@@ -95,7 +95,12 @@ class TemporaryShowyourworkRepository:
 
         # Parse options
         command = "showyourwork setup"
-        options = f"--quiet --action-spec={os.getenv('ACTION_SPEC')} "
+        options = "--quiet "
+        if os.getenv("ACTION_SPEC"):
+            options += f"--action-spec={os.getenv('ACTION_SPEC')} "
+        if os.getenv("ACTION_VERSION"):
+            options += f"--action-version={os.getenv('ACTION_VERSION')} "
+
         if self.cache:
             # Enable zenodo caching
             options += "--cache"
--- a/tests/integration/test_pr.py
+++ b/tests/integration/test_pr.py
@@ -4,7 +4,7 @@ import re
 
 import pytest
 import requests
-from helpers import TemporaryShowyourworkRepository
+from helpers import GITHUB_ORG, GITHUB_USER, TemporaryShowyourworkRepository
 
 from showyourwork import gitapi
 from showyourwork.subproc import get_stdout, parse_request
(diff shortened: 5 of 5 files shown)


HealthyPear:
OK at least now the test is using the correct code edits!

It is still failing though:
- either the current waiting settings are not enough
- or there is a problem even before...

HealthyPear:
by the way I tested the PR diff feature from the main branch and indeed it does not work, so this is not just a remote test failing, it is a real bug!

HealthyPear:
OK after some debugging I found that the diff fails to even get produced:

```
2026-03-03T11:12:12.1221913Z /bin/sh: 1: showyourwork: not found
2026-03-03T11:12:12.1295717Z ##[warning]Failed to generate diff with exec: /bin/sh: 1: showyourwork: not found
```

and this it trigger by the action source code

https://github.com/showyourwork/showyourwork-action/blob/f294fe2eb7275235596ff71c6de7b90d8e849fd6/src/publish.js

```javascript
// Build PDF diff
    if (core.getInput("build-diff-on-pull-request") == "true") {
      try {
        const LATEXDIFF_URL = core.getInput("latexdiff-url");
        const LATEXPAND_URL = core.getInput("latexpand-url");
        const LATEXDIFF_OPTS = core.getInput("latexdiff-options");
        const BASE_REF = github.context.payload.pull_request.base.sha;

        core.startGroup("Build article diff");
        shell.exec(`cp ${config["ms_pdf"]} .bkup.pdf`);

        // Download latexdiff and latexpand
        shell.exec(`wget ${LATEXDIFF_URL} && chmod +x latexdiff`);
        shell.exec(`wget ${LATEXPAND_URL} && chmod +x latexpand`);

        // Install perl and texlive packages:
        shell.exec(
          `sudo apt-get update -y && sudo apt-get install -y libalgorithm-diff-perl texlive-base`
        );

        // Checkout base version of ms.tex
        shell.exec(
          `./latexpand src/tex/${config["ms_name"]}.tex -o .flat_new.tex`
        );
        shell.exec(`git checkout ${BASE_REF} src/tex`);
        shell.exec(
          `./latexpand src/tex/${config["ms_name"]}.tex -o .flat_old.tex`
        );

        // Compute diff, and build
        shell.exec(
          `./latexdiff ${LATEXDIFF_OPTS} .flat_old.tex .flat_new.tex > tmp.tex`
        );
        // Patch for aastex to force exiting vertical mode before trying to render strikeout text.
        // If the first character of any paragraph or section is deleted, the diff will _not_
        // compile because the strikeout (\sout) command fails in vertical mode.
        shell.exec(
          `sed -i 's/\\\\RequirePackage\\[normalem\\]{ulem}/\\\\RequirePackage\\[normalem\\]{ulem} \\\\let\\\\oldsout\\\\sout \\\\renewcommand\\\\sout\\[1\\]{\\\\leavevmode\\\\oldsout{#1}}/' tmp.tex`
        );
        shell.exec(`mv tmp.tex src/tex/${config["ms_name"]}.tex`);
        shell.exec(`showyourwork build`);
        shell.exec(`cp ${config["ms_pdf"]} diff.pdf`);
        shell.exec(`cp .bkup.pdf ${config["ms_pdf"]}`);
        output.push("diff.pdf");
        core.endGroup();
      } catch (error) {
        // Raise warning, but don't fail the action
        shell.echo(`::warning ::Failed to generate diff with ${error.message}`);
      }
    }
```

I suspect that when the action calls showyourwork here it is not into the environment where showyourwork is installed 

I will open an issue / PR there first

HealthyPear:
See showyourwork/showyourwork-action#24

I do not know how much my changes here where actually needed, I suspect not and I will consider them debugging changes (we can save maybe the first commit, that I think is sensible anyway).

I am tempted to revert them by rebasing and force-pushing, therefore assuming that the test is well written, fix the behaviour of the action with a PR in the respective repo and see how that goes.

HealthyPear:
OK I can confirm that with showyourwork/showyourwork-action#25 (now merged in `main`) a test using my own account succeeded, I could download the diff pdf file

[diff.pdf](https://github.com/user-attachments/files/25717718/diff.pdf)

HealthyPear:
OK everything seems to be fixed now!

# It's the first time I see all green 🥹

PR #666, opened 2026-03-02
Fix PR links in the changelog

HealthyPear:
Right now in the main documentation if you go to any new changelog entry it will point to a PR search with filters

`is:open is:pr author:xxx`

where xxx is the number ID of the PR!

The main issue is caused by the fact that the link url contains `pulls` instead of `pull` so it looks for PRs made by an author but it's name is taken by the news fragment / PR id.

Apologies for not having realized this sooner 🙏


HealthyPear:
I can confirm that it now works as expected from the ReadTheDocs PR build!

PR #670, opened 2026-03-03
Allow for dependencies to be entire directories

### LINKED ISSUE #401: Allow for dependency to be a directory
In the documentation it says that dependencies can be files (see attached screenshot)

CleanShot 2023-09-20 at 10 15 36@2x

If I have a number of helper scripts in `utils` it would be great to only have to specify the `utils` directory and not enumerate each contained file. If I want to only be sensitive to select files I can of course specify them individually, but allowing for whole directories would be a great convenience.

HealthyPear:
#### config.py

- added `expand_dependency_directories()` function that recursively finds all files in a specified directory
- modified `get_upstream_dependencies()` to expand directory dependencies into individual files before recursing

#### preprocess.py

- imported the new expand_dependency_directories() function
- updated the dependency collection logic to expand any directory dependencies into individual files

#### docs

- updated the `dependencies` section to document that dependencies can now be directories
- added an example showing how to specify a whole directory as a dependency

#### tests

- added three new integration test classes scenarios I thought might happen (`test_directory_dependencies.py`):

  - single directory containing simple helper scripts, this is the new core feature
  - script depending on multiple directories simultaneously
  - individual files and directories mixed in the same dependency list

I did not make the tests remote as I think this is more of a internal structure thing

Closes #401

### DIFF
--- a/docs/changes/670.feature.rst
+++ b/docs/changes/670.feature.rst
@@ -1,4 +1,4 @@
 Dependencies in ``showyourwork.yml`` can be specified as entire directories.
-When a directory is specified (with a trailing slash, e.g., ``src/scripts/utils/``),
+When a directory is specified (e.g., ``src/scripts/utils/``, with or without a trailing slash),
 all files within that directory are treated as dependencies, eliminating the need to
 enumerate each helper script individually if they are too many to make it comfortable to do.
--- a/docs/changes/670.feature.rst
+++ b/docs/changes/670.feature.rst
@@ -1,0 +1,4 @@
+Dependencies in ``showyourwork.yml`` can be specified as entire directories.
+When a directory is specified (with a trailing slash, e.g., ``src/scripts/utils/``),
+all files within that directory are treated as dependencies, eliminating the need to
+enumerate each helper script individually if they are too many to make it comfortable to do.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -402,11 +402,15 @@ the path to a script (either a figure script or the TeX manuscript itself)
 relative to the repository root. Following each entry, provide a list of
 all files on which the script depends. These dependencies may either be
 static (such as helper scripts) or programmatically generated (such as
-datsets downloaded from Zenodo). In the latter case, instructions on how
+datasets downloaded from Zenodo). In the latter case, instructions on how
 to generate them must be provided elsewhere (either via the ``zenodo`` key
 below or via a custom ``rule`` in the ``Snakefile``). In both cases, changes
 to the dependency will result in a re-run of the section of the workflow that
 executes the script.
+
+Dependencies may be specified as individual files or as entire directories.
+When a directory is specified, all files within it will be treated as
+dependencies of the script.
 
 **Required:** no
 
@@ -421,6 +425,15 @@ the helper script ``utils/helper_script.py``:
   dependencies:
     src/scripts/my_figure.py:
       - src/scripts/utils/helper_script.py
+
+You can also specify a dependency on an entire directory. All files within
+the directory will be treated as dependencies:
+
+.. code-block:: yaml
+
+  dependencies:
+    src/scripts/my_figure.py:
+      - src/scripts/utils/
 
 You can also specify a dependency on a programmatically-generated file:
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -427,7 +427,7 @@ the helper script ``utils/helper_script.py``:
       - src/scripts/utils/helper_script.py
 
 You can also specify a dependency on an entire directory. All files within
-the directory will be treated as dependencies:
+the directory will be treated as dependencies (the trailing slash is optional):
 
 .. code-block:: yaml
 
--- a/tests/integration/test_directory_dependencies.py
+++ b/tests/integration/test_directory_dependencies.py
@@ -1,0 +1,250 @@
+"""Tests for directory dependencies feature.
+
+These tests verify that dependencies can be specified as entire directories,
+with all files within the directory treated as dependencies of a script.
+
+The tests are designed to be cross-platform compatible (Windows, macOS, Linux)
+by using pathlib for filesystem operations and forward slashes (/) in YAML
+configuration paths.
+"""
+
+import yaml
+from helpers import (
+    ShowyourworkRepositoryActions,
+    TemporaryShowyourworkRepository,
+)
+
+
+class TestDirectoryDependency(
+    TemporaryShowyourworkRepository, ShowyourworkRepositoryActions
+):
+    """
+    Test that a dependency can be specified as a directory.
+
+    When a directory is specified as a dependency in the showyourwork.yml,
+    all files within that directory should be treated as dependencies of the
+    dict key. This allows users to specify the entire
+    utils directory rather than listing each helper script individually.
+    """
+
+    # No need to test this on CI
+    local_build_only = True
+
+    def customize(self):
+        """Create figure script, helper scripts, and dependencies config."""
+
+        # Create helper scripts in a utils subdirectory
+        utils_dir = self.cwd / "src" / "scripts" / "utils"
+        utils_dir.mkdir(parents=True, exist_ok=True)
+
+        # Create first helper script
+        (utils_dir / "helper1.py").write_text(
+            "def helper1_function():\n    return 'Hello from helper1'\n"
+        )
+
+        # Create second helper script
+        (utils_dir / "helper2.py").write_text(
+            "def helper2_function():\n    return 'World from helper2'\n"
+        )
+
+        # Create a nested helper script
+        nested_dir = utils_dir / "nested"
+        nested_dir.mkdir(parents=True, exist_ok=True)
+        (nested_dir / "helper3.py").write_text(
+            "def helper3_function():\n    return 'Deep helper from helper3'\n"
+        )
+
+        # Create the main figure script that imports from utils
+        with open(self.cwd / "src" / "scripts" / "test_figure.py", "w") as f:
+            f.write(
+                "import sys\n"
+                "from pathlib import Path\n"
+                "sys.path.insert(0, str(Path.cwd() / 'src' / 'scripts'))\n"
+                "from utils.helper1 import helper1_function\n"
+                "from utils.helper2 import helper2_function\n"
+                "from utils.nested.helper3 import helper3_function\n"
+                "import matplotlib.pyplot as plt\n"
+                "import numpy as np\n"
+                "import paths\n"
+                "np.random.seed(0)\n"
+                "# Use the helper functions\n"
+                "msg1 = helper1_function()\n"
+                "msg2 = helper2_function()\n"
+                "msg3 = helper3_function()\n"
+                "# Create a simple figure\n"
+                "fig = plt.figure(figsize=(7, 6))\n"
+                "plt.plot([1, 2, 3, 4], [1, 4, 2, 3])\n"
+                "plt.title(f'{msg1} {msg2} {msg3}')\n"
+                "fig.savefig(paths.figures / 'test_figure.pdf', "
+                "bbox_inches='tight', dpi=300)\n"
+            )
+
+        # Add figure environment to the TeX file
+        self.add_figure_environment()
+
+        # Add dependency on the entire utils directory
+        # This is the key feature being tested - specifying a directory
+        # instead of individual files
+        config_path = self.cwd / "showyourwork.yml"
+        with open(config_path) as f:
+            config = yaml.safe_load(f)
+
+        config["dependencies"] = {
+            "src/scripts/test_figure.py": [
+                "src/scripts/utils/",  # Specify the entire utils directory
+            ],
+        }
+
+        with open(config_path, "w") as f:
+            yaml.dump(config, f, default_flow_style=False)
+
+    def check_build(self):
+        """Verify that the figure was generated successfully."""
+        # The figure should have been generated
+        assert (
+            self.cwd / "src" / "tex" / "figures" / "test_figure.pdf"
+        ).exists(), "Figure was not generated despite having directory dependency"
+
+
+class TestDirectoryDependencyMultiple(
+    TemporaryShowyourworkRepository, ShowyourworkRepositoryActions
+):
+    """
+    Test that multiple directory dependencies work correctly.
+
+    This verifies that a script can depend on multiple directories,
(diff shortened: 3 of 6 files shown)


HealthyPear:
We apparently hit this open upstream issue

https://github.com/binaryornot/binaryornot/issues/634

see also 

https://github.com/cookiecutter/cookiecutter/pull/2196

HealthyPear:
@nstarman not sure if you still use showyourwork, but if yes, perhaps you can test this with your projects to see if it behaves like you would expect?

HealthyPear:
@vandalt can you test/review this? nobody else answered 😅

docs/config.rst
```diff
@@ -422,6 +426,15 @@ the helper script ``utils/helper_script.py``:
     src/scripts/my_figure.py:
       - src/scripts/utils/helper_script.py
 
+You can also specify a dependency on an entire directory. All files within
+the directory will be treated as dependencies:
```
vandalt:
```suggestion
the directory will be treated as dependencies (the trailing slash is optional):
```

docs/changes/670.feature.rst
```diff
@@ -0,0 +1,4 @@
+Dependencies in ``showyourwork.yml`` can be specified as entire directories.
+When a directory is specified (with a trailing slash, e.g., ``src/scripts/utils/``),
```
vandalt:
```suggestion
When a directory is specified (e.g., ``src/scripts/utils/``, with or without a trailing slash),
```

vandalt: From what I understand, the trailing slash is optional since you use `is_dir()`. Is this correct?

Otherwise I do think it would be preferable to avoid requiring the trailing slash.

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+            config = yaml.safe_load(f)
+
+        config["dependencies"] = {
+            "src/scripts/test_figure.py": [
+                "src/scripts/utils/",  # Specify the entire utils directory
+            ],
+        }
+
+        with open(config_path, "w") as f:
+            yaml.dump(config, f, default_flow_style=False)
```
vandalt: Replace with `edit_yaml()` (see [this example](https://github.com/showyourwork/showyourwork/blob/aab40cf476879364a4519df7abd23ed041cd3d8d/tests/integration/test_cache.py#L37-L41))

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+
+        config["dependencies"] = {
+            "src/scripts/test_figure.py": [
+                "src/scripts/utils/",
+                "src/scripts/config/",  # Multiple directories
+            ],
+        }
+
+        with open(config_path, "w") as f:
+            yaml.dump(config, f, default_flow_style=False)
```
vandalt: Replace with `edit_yaml()` (see [this example](https://github.com/showyourwork/showyourwork/blob/aab40cf476879364a4519df7abd23ed041cd3d8d/tests/integration/test_cache.py#L37-L41))

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+
+        config["dependencies"] = {
+            "src/scripts/test_figure.py": [
+                "src/scripts/standalone_helper.py",  # Individual file
+                "src/scripts/utils/",  # Directory
+            ],
+        }
+
+        with open(config_path, "w") as f:
+            yaml.dump(config, f, default_flow_style=False)
```
vandalt: Replace with `edit_yaml()` (see [this example](https://github.com/showyourwork/showyourwork/blob/aab40cf476879364a4519df7abd23ed041cd3d8d/tests/integration/test_cache.py#L37-L41))

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+                "msg1 = helper1_function()\n"
+                "msg2 = helper2_function()\n"
+                "msg3 = helper3_function()\n"
+                "# Create a simple figure\n"
+                "fig = plt.figure(figsize=(7, 6))\n"
+                "plt.plot([1, 2, 3, 4], [1, 4, 2, 3])\n"
+                "plt.title(f'{msg1} {msg2} {msg3}')\n"
+                "fig.savefig(paths.figures / 'test_figure.pdf', "
+                "bbox_inches='tight', dpi=300)\n"
+            )
```
vandalt, line 84: I would be tempted to move this to a separate method (maybe along with the config file edit below) to keep the customize method more readable

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+    and all files within each directory are properly tracked.
+    """
+
+    # No need to test this on CI
+    local_build_only = True
+
+    def customize(self):
+        """Create multiple directories with helper scripts."""
+
+        # Create first utils directory
```
vandalt: I would be tempted to move this to a separate method (maybe along with the config file edit below) to keep the customize method more readable

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+                "import numpy as np\n"
+                "import paths\n"
+                "np.random.seed(0)\n"
+                "msg1 = standalone()\n"
+                "msg2 = from_utils()\n"
+                "fig = plt.figure(figsize=(7, 6))\n"
+                "plt.plot([1, 2, 3])\n"
+                "fig.savefig(paths.figures / 'test_figure.pdf', "
+                "bbox_inches='tight', dpi=300)\n"
+            )
```
vandalt: I would be tempted to move this to a separate method (maybe along with the config file edit below) to keep the customize method more readable

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+"""
+
+import yaml
+from helpers import (
+    ShowyourworkRepositoryActions,
+    TemporaryShowyourworkRepository,
+)
+
+
+class TestDirectoryDependency(
```
vandalt, line 15: Since these are integration tests, I'm wondering how much we need to isolate the three use cases. I could see one big test that has all three cases covered (two directory dependency and one file).

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+        config["dependencies"] = {
+            "src/scripts/test_figure.py": [
+                "src/scripts/utils/",  # Specify the entire utils directory
+            ],
+        }
+
+        with open(config_path, "w") as f:
+            yaml.dump(config, f, default_flow_style=False)
+
+    def check_build(self):
```
vandalt: I think it would be good to also modify one of the script dependencies, re-run the workflow and make sure the figure was re-generated (by saving the timestamp of the output file before and after or capturing the output for example.

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+            "src/scripts/test_figure.py": [
+                "src/scripts/utils/",
+                "src/scripts/config/",  # Multiple directories
+            ],
+        }
+
+        with open(config_path, "w") as f:
+            yaml.dump(config, f, default_flow_style=False)
+
+    def check_build(self):
```
vandalt, line 104: I think it would be good to also modify one of the script dependencies, re-run the workflow and make sure the figure was re-generated (by saving the timestamp of the output file before and after or capturing the output for example.

tests/integration/test_directory_dependencies.py
```diff
@@ -0,0 +1,250 @@
...
+            "src/scripts/test_figure.py": [
+                "src/scripts/standalone_helper.py",  # Individual file
+                "src/scripts/utils/",  # Directory
+            ],
+        }
+
+        with open(config_path, "w") as f:
+            yaml.dump(config, f, default_flow_style=False)
+
+    def check_build(self):
```
vandalt: I think it would be good to also modify one of the script dependencies, re-run the workflow and make sure the figure was re-generated (by saving the timestamp of the output file before and after or capturing the output for example.

vandalt (changes requested):
Looks good to me overall, but maybe the test customization structure could be simplified a bit and the `check_build()` should probably test that modifying the dependency triggers a re-run.

Also, as one of my comments mentions, maybe we could merge all three use cases in a single integration test. The specifics of the dependency discovery could actually be covered by unit tests that target the functions in `config.py` and `preprocess.py` if we want to cover many use-cases separately.

I left specific comments in the code directly for all three methods.

tests/integration/test_directory_dependencies.py
vandalt: I think a simple way to do this would be like what is done with `test_cache.py` where the same script is generated with a different random seed in some random generated. Maybe we could even re-use some code from there

vandalt:
Testing locally, this seems to work as expected on a small test repo :)

docs/config.rst
HealthyPear: fixed

docs/changes/670.feature.rst
HealthyPear: yes correct fixed

tests/integration/test_directory_dependencies.py
HealthyPear: done everywhere thx!

tests/integration/test_directory_dependencies.py
HealthyPear: I refactored this and the edit of the dependencies section too

tests/integration/test_directory_dependencies.py
HealthyPear: done

tests/integration/test_directory_dependencies.py
HealthyPear: done!

tests/integration/test_directory_dependencies.py
HealthyPear: I am now checking for a rebuild if the image script gets modified and that part has now random seed support

I am then checking the mtime of the generated image before and after

HealthyPear:
last commit did:

- merge 3 use cases inot a single test
- use edit_yaml everywhere
- refactored customized into separate methods for better readability
- confirm the dependency change triggers a rebuild
- added seed parameter to _create_figure_script()

vandalt (approved):
Looks good to me! And the test is still passing locally.

HealthyPear:
Damn it seems that the latest changes after the first review broke some integration local tests first only on windows then also ubuntu

it seems to be something limited to the remote GitHub runners because the patch is looking into paths that resolve outside of the repo root path for some reason

vandalt:
The error occurs in `test_hooks.py`, which was added as part of #688 (by me :upside_down_face:). I confirm that cloning this PR and rebasing on main, I can reproduce the `test_hooks.py` failure.

It was an error in how I had configured the dependencies config in the test. I was using a string instead of a list so the iteration in `get_upstream_dependencies()` was looping over a string instead of a list.

See #706 for a hotfix.

Sorry about that @HealthyPear

HealthyPear:
Apparently the fix in #706 was not sufficient: `TestInclude::test_local` is now failing for the same reason, i.e. some paths are resolved outside of the test project repo

I have no idea why this happens only on the remote CI

I guess what we could do is filter out the non-sensical paths like this for example, 

```python
files = []
for file in xml_tree.findall("INPUT"):
    resolved = (paths.user().tex / file.text).resolve()
    try:
        rel = resolved.relative_to(paths.user().repo)
        files.append(rel.as_posix())
    except ValueError:
        # Skip files that resolve outside the repository root
        pass
```

and analogously for graphics

```python
graphics = []
for graphic in figure.findall("GRAPHICS"):
    resolved = (paths.user().tex / graphicspath / graphic.text).resolve()
    try:
        rel = resolved.relative_to(paths.user().repo)
        graphics.append(rel.as_posix())
    except ValueError:
        pass 
```

and free-floating graphics 

```python
free_floating_graphics = []
for graphic in xml_tree.findall("GRAPHICS"):
    resolved = (paths.user().tex / graphicspath / graphic.text).resolve()
    try:
        rel = resolved.relative_to(paths.user().repo)
        free_floating_graphics.append(rel.as_posix())
    except ValueError:
        pass
free_floating_graphics += unlabeled_graphics
```

but I'd still like to understand why this happens

HealthyPear:
Actually the problem might be the use of `.resolve()`: I am used to use it because it's the most general thing to do usually, but in this case it might make more harm than good

the paths of the dependencies should be already absolute from the joining and even if a user has a ../ in the config that should be correctly dealt with by the `.relative_to()` (and if this does not work I would prefer to et the thing crash as due to an authoring error rather than silently resolving to a system file.

vandalt:
@HealthyPear I think the issue for `test_include` is the same as it was for `test_hooks`... We need to repce `"src/scripts/test_figure.py": "src/data/test_data.npz"` by `"src/scripts/test_figure.py": ["src/data/test_data.npz"]`.

Do you want me to submit a PR? I think it would be OK if you do it as part of this one although not directly related to the changes you are making.


Sorry again!

HealthyPear:
> @HealthyPear I think the issue for `test_include` is the same as it was for `test_hooks`... We need to repce `"src/scripts/test_figure.py": "src/data/test_data.npz"` by `"src/scripts/test_figure.py": ["src/data/test_data.npz"]`.
> 
> Do you want me to submit a PR? I think it would be OK if you do it as part of this one although not directly related to the changes you are making.
> 
> Sorry again!

I can try after this last change, I really think resolving symlinks is too dangerous anyway

vandalt:
> I can try after this last change, I really think resolving symlinks is too dangerous anyway

Sounds good.

Otherwise I can also make a dedicated PR for this but it does need to be fixed. Otherwise the tests does not check the correct path (it checks each char in the string as a file).

HealthyPear:
OK that was it @vandalt thanks!

Could you give a last review?

vandalt:
Sorry I didn't have time to review this. But will definitely be testing this soon-ish in my actual usage :)

Thanks!

HealthyPear:
> Sorry I didn't have time to review this. But will definitely be testing this soon-ish in my actual usage :)
> 
> Thanks!

no worries, I tested it before the last fixes and it was already working well for me

though after merging I see that it is failing in the remote tests, even though it was succeeding here (after the fix you suggested were the dependencies has to be made a list) - I will check in the next days

PR #678, opened 2026-03-09
Unpin chardet now that binaryornot fixed the issue upstream

vandalt:
Based on [the latest binaryornot releases](https://github.com/binaryornot/binaryornot/releases), the issue with chardet is fixed (I think in 0.6.0 chardet is no longer used at all). This means we can unpin `chardet` on our end.

I tested `test_default.py` with `binaryornot==0.4.4` to confirm the tests were failing, then with `binaryornot==0.6.0` to confirm the issue was gone.

I also added an import test to make sure the weird `requests` dependency warning was gone.

### DIFF
--- a/tests/unit/test_import.py
+++ b/tests/unit/test_import.py
@@ -1,0 +1,22 @@
+import sys
+
+import pytest
+
+
+def test_requests_import_no_dependency_warning():
+    """Test that importing requests doesn't produce RequestsDependencyWarning."""
+    # Remove requests from sys.modules to force reimport
+    modules_to_remove = [key for key in sys.modules if key.startswith("requests")]
+    for module in modules_to_remove:
+        del sys.modules[module]
+
+    # Use pytest's warning recorder to capture all warnings
+    with pytest.warns() as warning_list:
+        import requests  # noqa: F401
+
+    # Check if any RequestsDependencyWarning was raised
+    for warning in warning_list:
+        if "RequestsDependencyWarning" in str(warning.category):
+            pytest.fail(f"RequestsDependencyWarning detected: {warning.message}")
+        if "RequestsDependencyWarning" in str(warning.message):
+            pytest.fail(f"RequestsDependencyWarning detected: {warning.message}")
--- a/tests/unit/test_import.py
+++ b/tests/unit/test_import.py
@@ -1,4 +1,5 @@
 import sys
+import warnings
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_import.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_requests_import_no_dependency_warning():
+    """Test that importing requests doesn't produce RequestsDependencyWarning."""
+    # Remove requests from sys.modules to force reimport
+    modules_to_remove = [key for key in sys.modules if key.startswith("requests")]
+    for module in modules_to_remove:
+        del sys.modules[module]
+
+    # Use pytest's warning recorder to capture all warnings
+    with pytest.warns() as warning_list:
```
HealthyPear: What about doing this instead?

https://docs.pytest.org/en/stable/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

> To ensure that no warnings are emitted, use:
> 
```python
def test_warning():
   with warnings.catch_warnings():
       warnings.simplefilter("error")
       ...
```

vandalt: This simplifies the code a bit but feels less specific as it would error out on any warning. What I'm doing is closer to the section right after: https://docs.pytest.org/en/stable/how-to/capture-warnings.html#custom-failure-messages.

I prefer the current approach which specifically check for a warning type I think. We could also filter based on the warning message with `        warnings.filterwarnings("error", message=".*supported version.*")`, but filtering based on type requires importing `requests`, which makes the test circular.

Let me know what you prefer.

HealthyPear: I was suggesting it because currently your unit test is currently failing because there a re no warnings, so probably a combination of the two might catch all cases?

HealthyPear (changes requested):

tests/unit/test_import.py
vandalt: Oops I had missed that. I updated it to catch warnings with `warnings` directly, but keep the custom failure logic to be as specific as possible about the warning type.

I'll let you mark as resolved if you are OK with the fix.

HealthyPear:
sorry I must have removed my own comment by mistake!

I was suggesting to do (the opposite) of what they suggest i the link you posted

```
if not warning_list:
         # make test pass
```

vandalt:
> I was suggesting to do (the opposite) of what they suggest i the link you posted

But this would fail for _any_ warning, correct? In that case the `simplefilter` approach you suggested is indeed better. The only advantage of what I'm doing right now is we can test specifically for the `RequestsDependencyWarning`. If we want to test for _any_ warning then I can switch to `simplefilter`.

vandalt:
@HealthyPear should I update to match all warnings or keep the PR as-is (I was not sure how to interpret the thumbs up reaction haha)?

Thanks!

HealthyPear:
> @HealthyPear should I update to match all warnings or keep the PR as-is (I was not sure how to interpret the thumbs up reaction haha)?
> 
> Thanks!

no no it's fine like you did I just wanted to cover the case of no warnings

thank you!

HealthyPear (approved):

Issue #434, opened 2024-01-23
Zenodo sandbox entries named after jobs instead of files


afarah18:
Hello, hoping I could get some advice on using dynamic datasets with zenodo sandbox. 

**Summary**
I've set up zenodo sandbox caching for two of my intermediate results, one of which is very expensive to generate. While it seems like my zenodo sandbox authentication is working, and the sandbox entries are created, SYW can never find the correct files in those entries. I suspect it might be because the sandbox entries are named according to the job name as specified by the Snakefile, but SYW looks for the file names. I can't tell if I'm using one where I should be using the other, or if there is a bug. 

**Some more details**
My Snakefile contains the following:
```python
rule datagen:
    output:
        directory("src/data/gw_data")
    cache:
        True
    input:
        "src/data/optimal_snr_aplus_design_O5.h5"
    script:
        "src/scripts/data_generation.py"
rule nonparinference:
    output:
        "src/data/mcmc_nonparametric.nc4"
    cache:
        True
    input:
        "src/data/gw_data"
    script:
        "src/scripts/nonparametric_inference.py"
rule nonparplots:
    output:
        "src/tex/figures/O5_GP.pdf"
    input:
        "src/data/mcmc_nonparametric.nc4"
    script:
        "src/scripts/nonparametric_twopanel.py"
```

  When I run `showyourwork build`, after `showyourwork clean` I get 
```User authentication for 10.5072/zenodo.23379 is valid.
...
Running user rule datagen...
Searching remote file cache: src/data/gw_data...
```
which tells me that it is successfully going to the zenodo sandbox entry. However, even after I have run this several times, I always get 
```
File not found on remote cache. See logs for details.
Running rule from scratch...
Caching output file on remote: src/data/gw_data...
```
This happens for both of my intermediate results. When I look at the sandbox entry that was created, it shows files with the name of the jobs I specified in the Snakefile, not the outputs of those jobs (see below screenshot) 
<img width="486" alt="image" src="https://github.com/showyourwork/showyourwork/assets/13318208/4fc6ba7a-6e5c-4ce5-82d6-037bee8740d0">

The `showyourwork.log` says this:
```
Attempting to access Zenodo Sandbox deposit with DOI 10.5072/zenodo.23379...
Searching for file `datagen` with hash `b324a29d7f46bbb604af71a59d0cee465693186cc3c2c674883ee525ad13750e`...
File not found on remote cache. See logs for details.
list indices must be integers or slices, not str
[...]

PR #680, opened 2026-03-12
Fix new draft creation after a Zenodo deposit is frozen

### LINKED ISSUE #638: [Bug]: No new draft created for Zenodo after the deposit is frozen
### Preliminary checks

The docs mention that after freezing a zenodo deposit, a new draft will be created and used. This does not seem to happen in the testing I've done, as mentioned in #635 (which documents this as a known issue).

### To Reproduce

1. Run `showyourwork build` with Zenodo integration
2. Run `showyourwork cache freeze`
3. Change something in the output
4. Re-run `showyourwork build`
5. The file is generated, but not uploaded anywhere.

### Logs

_No response_

### Setup information

```python
```

vandalt:
This PR fixes a few problems with how frozen deposits are handled:

- Update the `TestCacheFreeze` unit test to fail when no new draft is created after freezing the Zenodo deposit
- Fix a `KeyError` that prevented showyourwork from finding older published versions of a file when a new draft existed
- Fix the detection of an existing draft: even when a record has been submitted, its `latest_draft` link still exists. We need to check the `submitted` JSON field to see if a draft exists. This caused showyourwork to always skip the creation of a new version.

The 3rd bullet point is fixed in `upload_file()`  but not in the other two instances where the `newversion` API is used so I will mark as a draft for now.

Fixes #638.


vandalt:
@HealthyPear I think this is ready for review. Everything works locally when I run the tests. However the unit tests don't seem to have access to Zenodo. I tried to add SANDBOX_TOKEN to the unit tests workflow but that does not fix the issue... Maybe it needs to be done on the main branch?

Thanks!

HealthyPear:
> @HealthyPear I think this is ready for review. Everything works locally when I run the tests. However the unit tests don't seem to have access to Zenodo. I tried to add SANDBOX_TOKEN to the unit tests workflow but that does not fix the issue... Maybe it needs to be done on the main branch?
> 
> Thanks!

mmm weird...didn't they pass all in the main Zenodo PR you did sometime ago?

HealthyPear:
they were passing all there

https://github.com/showyourwork/showyourwork/actions/runs/22400674392

HealthyPear:
is it perhaps the first time we call `self.access_token` from the unit tests? maybe that workflow has different settings from the repo point of view?

vandalt:
> mmm weird...didn't they pass all in the main Zenodo PR you did sometime ago?

Yes, the ones that are failing were added as part of this PR.

Is it possible that the `SANDBOX_TOKEN` secret is added to the repos that get created for remote integration tests, but is not added to this repo? If that is not the issue I'll have to dig a bit more to understand what is happening.

HealthyPear:
> > mmm weird...didn't they pass all in the main Zenodo PR you did sometime ago?
> 
> Yes, the ones that are failing were added as part of this PR.
> 

ok this at least confirms that it is something related to thee changes and should not be propagated to main or other branches

> Is it possible that the `SANDBOX_TOKEN` secret is added to the repos that get created for remote integration tests, but is not added to this repo? If that is not the issue I'll have to dig a bit more to understand what is happening.

mmm technically it should be a repository secret, so I expect it to be available from everywhere in any workflow

HealthyPear:
I added the safe to test label, I am curious to see if also the remote tests fail, in that case maybe there is something wrong with the token itself?

HealthyPear:
the remote tests completed all succesfully, so it is definitely something going on between your changes and exclusively the unit test workflow or the way your are interfacing with the Zenodo API via your new test versions

vandalt:
> the remote tests completed all succesfully, so it is definitely something going on between your changes and exclusively the unit test workflow or the way your are interfacing with the Zenodo API via your new test versions

But the remote tests use the token only on the "test" repos that get created inside the test correct? Is it possible that these repos have access to the `SANDBOX_TOKEN` but not the main `showyourwork` repo.

vandalt:
The latest commit fixes #434 completely.

@HealthyPear let me know your thoughts on [my previous comment](https://github.com/showyourwork/showyourwork/pull/680#issuecomment-4055674182) and I'll look for a fix once I have the confirmation that the token is indeed accessible from this repo.

Thanks!

vandalt:
I added the `@require_access_token` wrapper to `_create` and I confirm that the token is not found by the tests. I'm not sure how to enable access to the `SANDBOX_TOKEN`. I already added it to the workflow file ([here](https://github.com/showyourwork/showyourwork/pull/680/changes#diff-c91058a32040621ac54d6ea4194b302a420b7849f6e8561920d23a0e97acf69aR53))

```
      - name: Run tests
        shell: bash -l {0}
        run: |
          python -m pytest tests/unit
        env:
          SANDBOX_TOKEN: ${{ secrets.SANDBOX_TOKEN }}
```

vandalt:
Looks like I broke the local cache optimization test too. I'll look into that next week

vandalt:
> I added the `@require_access_token` wrapper to `_create` and I confirm that the token is not found by the tests. I'm not sure how to enable access to the `SANDBOX_TOKEN`. I already added it to the workflow file ([here](https://github.com/showyourwork/showyourwork/pull/680/changes#diff-c91058a32040621ac54d6ea4194b302a420b7849f6e8561920d23a0e97acf69aR53))
> 
> ```
>       - name: Run tests
>         shell: bash -l {0}
>         run: |
>           python -m pytest tests/unit
>         env:
>           SANDBOX_TOKEN: ${{ secrets.SANDBOX_TOKEN }}
> ```

@HealthyPear since we have the confirmation that the tests fail due to sandbox token access, I think we have two options:

1. Fix the root issue by making sure the tests running on this repo (and not just on repos created by the integration tests) have access to the `SANDBOX_TOKEN`
2. Add a "skip if" to the tests to skip tests that require the sandbox token when none is available.

I think adding option 2 would be a good idea in any case so that users can run the tests without setting up zenodo. But I would tend to favor option 1 on the GitHub actions.

vandalt:
> Looks like I broke the local cache optimization test too. I'll look into that next week

Actually, they are broken on main too so I'll open a separate issue for that (it might be due to #681 too). @HealthyPear could you mark the PR safe for remote tests to make sure nothing else is failing in integration?

Thanks!

EDIT: Downgrading snakemake to 9.16.3 fixes the tests so I think this is the root cause.

HealthyPear:
> > Looks like I broke the local cache optimization test too. I'll look into that next week
> 
> Actually, they are broken on main too so I'll open a separate issue for that (it might be due to #681 too).

very likely

> @HealthyPear could you mark the PR safe for remote tests to make sure nothing else is failing in integration?
> 
> Thanks!
> 

done

> EDIT: Downgrading snakemake to 9.16.3 fixes the tests so I think this is the root cause.

9.17.0 also breaks it?

HealthyPear:
> 1. Fix the root issue by making sure the tests running on this repo (and not just on repos created by the integration tests) have access to the `SANDBOX_TOKEN`
> 2. Add a "skip if" to the tests to skip tests that require the sandbox token when none is available.
> 
> I think adding option 2 would be a good idea in any case so that users can run the tests without setting up zenodo. But I would tend to favor option 1 on the GitHub actions.

actually both

option 2 should come with a reason (I think it is possible from pytest to explain why the test was skipped and it should point to our zenodo docs)

vandalt:
> actually both
>
> option 2 should come with a reason (I think it is possible from pytest to explain why the test was skipped and it should point to our zenodo docs)

That works.

Regarding option 1, let me know if there is anything I can do to help on my side. I could try running the tests on my fork to debug this and set the token myself to debug the issue, but I cannot edit the token on the main showyourwork repo

Regarding option 2, I will add it with a reason and with a CLI flag to require zenodo access (i.e. never skip the zenodo tests). I think this CLI flag could be used for GitHub actions.

vandalt:
@HealthyPear when I ran it on my fork, I got a failure at first without the token set, but after adding `SANDBOX_TOKEN` as a repository secret everything works (Settings -> Secrets and Variables > Actions, then adding the token under Repository secrets). Here is a link to the workflow run: https://github.com/vandalt/showyourwork/actions/runs/23160000022/job/67284999805.

Are you able to set or update the token on this repo to see if it helps?

HealthyPear:
> @HealthyPear when I ran it on my fork, I got a failure at first without the token set, but after adding `SANDBOX_TOKEN` as a repository secret everything works (Settings -> Secrets and Variables > Actions, then adding the token under Repository secrets). Here is a link to the workflow run: https://github.com/vandalt/showyourwork/actions/runs/23160000022/job/67284999805.
> 
> Are you able to set or update the token on this repo to see if it helps?

I recently got my permissions bumped up, so I should be able to do practically everything

we already had a SANDBOX_TOKEN **repository** secret (which should be the same we need for the unit tests?)

but I can see there is a SANDBOX_TOKEN **organization** secret

perhaps I can re-create a SANDBOX_TOKEN (for now it is associated to my zenodo account, but we should probably make a showyourwork account unless one of the maintainers already did it) and use the same new updated token for both

vandalt:
> we already had a SANDBOX_TOKEN repository secret (which should be the same we need for the unit tests?)

Maybe it's expired or does not have write access? Just guessing here.

> we should probably make a showyourwork account

Sounds like a good idea.

HealthyPear:
> > we already had a SANDBOX_TOKEN repository secret (which should be the same we need for the unit tests?)
> 
> Maybe it's expired or does not have write access? Just guessing here.
> 

ok I generated a new token and used it for both repo and org

vandalt:
Still failing... I really don't understand the difference between why the token is not accessible here but it is on [my fork](https://github.com/vandalt/showyourwork/actions/runs/23160000022/job/67284999805)...

vandalt:
Ah it seems secrets are not passed to workflows running from a fork: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow.

I suggest we skip the tests that require the secret if the token is not found, and use the integration tests to assess if Zenodo access works on pull requests.

I think the Zenodo unit tests are still useful for development (quicker than integration) but it's not the end of the world if they don't work on actions.

HealthyPear:
> Ah it seems secrets are not passed to workflows running from a fork: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow.
> 

aaaahh right I forgot about that

but wait the secret is not from your fork no? or github thinks it is because you (aka your fork's branch triggered the workflow?)

what about all other previous PRs using secrets that you worked on - wasn't that the same? I am so confused now ahah

I would imagine I would have got a banner from github asking me to approve the workflow if there was this kind of security issue...but again I just got yesterday full powers, so I did not check all existing settings

> I suggest we skip the tests that require the secret if the token is not found, and use the integration tests to assess if Zenodo access works on pull requests.

deal

> I think the Zenodo unit tests are still useful for development (quicker than integration) but it's not the end of the world if they don't work on actions.

I guess, but when this PR is merged, than the workflow at least on main should be triggered from this repo and then those test should run, am I correct?

vandalt:
@HealthyPear I just set up the unit tests to skip the Zenodo tests if no token is available and they now pass.

I also pinned `snakemake<9.17` (`9.17.0` also failed due to #681) so that remote tests would pass to ensure the PR did not break anything. If you want to mark it as `safe to test`, I think it would then be ready for final review!

vandalt:
> what about all other previous PRs using secrets that you worked on - wasn't that the same? I am so confused now ahah

Previous PRs used secrets in integration tests only, which is OK because the repos are created by the showyourwork org (and this is why `safe to test` is required).

> I guess, but when this PR is merged, than the workflow at least on main should be triggered from this repo and then those test should run, am I correct?

Yes, triggering the workflow on main (or any branch from `showyourwork/showyourwork`) should work.

PR #682, opened 2026-03-16
Remove deprecated `cache_mode` argument from snakemake patches

### LINKED ISSUE #681: [Bug]: Cache tests are failing after snakemake update
### Preliminary checks

This snakemake PR removed the `cache_mode` argument from the `fetch()` and `store()` caching methods: https://github.com/snakemake/snakemake/pull/4085. This breaks things on our side. I'm working on the fix.

Hopefully I'll just have to remove the argument in `showyourwork` patches.

### To Reproduce

`python -m pytest tests/integration/test_cache.py`

### Logs

_No response_

### Setup information

_No response_

vandalt:
Fixes #681 by updating the snakemake patches to follow the updated API from `snakemake>=9.17.0`, more specifically the changes in https://github.com/snakemake/snakemake/pull/4085.

I would probably merge this after #680 since it allows us to unpin Snakemake.

### DIFF
--- a/docs/changes/682.bugfix.rst
+++ b/docs/changes/682.bugfix.rst
@@ -1,0 +1,1 @@
+Fix the Snakemake caching by removing the deprecated `cache_mode` argument from the cache API.
--- a/docs/changes/682.bugfix.rst
+++ b/docs/changes/682.bugfix.rst
@@ -1,1 +1,1 @@
-Fix the Snakemake caching by removing the deprecated `cache_mode` argument from the cache API.
+Fix the caching for Snakemake 9.17.0 and require `snakemake>=9.17.0`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     "click",
     "cookiecutter",
     "packaging",
-    "snakemake<9.17",
+    "snakemake>=9.17.0",
     "pulp",
 ]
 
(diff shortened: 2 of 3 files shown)


docs/changes/682.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix the Snakemake caching by removing the deprecated `cache_mode` argument from the cache API.
```
HealthyPear: just a small formality: news fragments should be written in a way that "speaks" to the end user (like peasants that listen to the "towncrier" 😄 ) instead of the language you would use for commits

for the end user in this case the only relevant change is that with this PR the version of  snakemake that will be installed is >=9.17

HealthyPear (changes requested):

docs/changes/682.bugfix.rst
vandalt: Ah I see, didn't know that was what towncrier stands for haha.

I updated the message. I think it's still worth mentioning that the caching is fixed, since technically it's possible someone updated snakemake, then encountered the issue, and this would fix it. But I kept the message short and remove info about internal details.

Let me know if I should change it again, otherwise feel free to mark as resolved.

vandalt:
I just rebased on `main` to fix the conflict with #680. This should be ready for merge I think @HealthyPear!

vandalt:
The failing test is about a conda HTTP error. I guess this can be safely ignored?

HealthyPear (approved):
Thanks!

PR #684, opened 2026-03-19
Add the possibility to let the output through when debugging integration tests

vandalt:
In the integration tests, `showyourwork` is called through the `get_stdout()` function, which always calls `subprocess.run()` with `capture_output=True`. This is an issue when trying to debug `showyourwork` code with `ipdb` during the integration tests.

This PR makes `capture_output` an argument to `get_stdout()` (`True` by default) and uses the existing `self.debug` flag to determine whether the output should be captured when running the integration tests. I also added a mention of this in the integration tests docs.

### DIFF
--- a/docs/integration_tests.rst
+++ b/docs/integration_tests.rst
@@ -96,6 +96,18 @@ the output from the Python script.
 That's it! The parent class (``TemporaryShowyourworkRepository``) takes care
 of the rest.
 
+Debugging within the tests
+--------------------------
+
+By default, the integration tests capture the output when running the ``showyourwork`` command.
+This will cause problems when trying to debug parts of the main ``showyourwork`` code within the tests.
+To do so, you will need to enable debug mode to let the output pass through:
+
+.. code-block:: bash
+
+    DEBUG=1 python -m pytest tests/integration/test_default.py -s
+
+
 Remote tests
 ------------
 
--- a/src/showyourwork/subproc.py
+++ b/src/showyourwork/subproc.py
@@ -22,7 +22,13 @@ def process_run_result(code, stdout, stderr):
 
 
 def get_stdout(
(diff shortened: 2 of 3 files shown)


vandalt:
@HealthyPear I think no changelog is needed here? Or should there be a "maintenance" one?

Also, since this changes the integration tests, it's probably worth marking as `safe to test` to run the remote tests.

HealthyPear:
> @HealthyPear I think no changelog is needed here? Or should there be a "maintenance" one?

no since it is a purely internal testing feature, I do not see a need to add an entry to the changelog

I see that you already updated the docs so that is perfectly fine 👍 

> Also, since this changes the integration tests, it's probably worth marking as `safe to test` to run the remote tests.

will do!

vandalt:
The remote tests failure all seem to be Conda HTTP errors. I'll let you decide if you want to re-run them or merge but the failures do not seem related to this PR.

```
FAILED tests/integration/test_default.py::TestDefaultDry::test_remote - Exception: [test-default-dry] GitHub Actions workflow terminated with status failure.
For details, see https://github.com/showyourwork/test-default-dry/actions/runs/23340551594
FAILED tests/integration/test_latex.py::TestMatplotlibLatex::test_remote - Exception: [test-matplotlib-latex] GitHub Actions workflow terminated with status failure.
For details, see https://github.com/showyourwork/test-matplotlib-latex/actions/runs/23340552787
FAILED tests/integration/test_zenodo.py::TestZenodo::test_remote - Exception: [test-zenodo] GitHub Actions workflow terminated with status failure.
For details, see https://github.com/showyourwork/test-zenodo/actions/runs/23340558340
```

HealthyPear:
Everything is fine now! Thanks!

PR #685, opened 2026-03-19
Add a `showyourwork cache reserve` command to pre-reserve a DOI from the main Zenodo service

### LINKED ISSUE #402: Add an option for reserving a DOI from Zenodo
When you're near to finalizing a paper there's no way to put in the Zenodo information without publishing the datasets, transferring from the Sandbox. However it's possible to reserve a DOI. Perhaps `showyourwork cache reserve` could be added to reserve a DOI before actually publishing.
Thanks!

vandalt:
This PR adds a `showyourwork cache reserve` command which will create a draft on the main zenodo.org service without pushing anything to it, and register the DOI in `zenodo.yaml`.

Then, the `showyourwork cache publish` command will use this existing DOI from the config file.

If a DOI already exists, an error is raised to avoid any ambiguity.

Fixes #402.

Marking as draft for now as I need to add tests.

### DIFF
--- a/src/showyourwork/cli/commands/zenodo.py
+++ b/src/showyourwork/cli/commands/zenodo.py
@@ -6,6 +6,41 @@ from ...git import get_repo_branch
 from ...zenodo import Zenodo
 
 
+def zenodo_reserve(branch):
+    """Reserve a DOI on Zenodo
+
+    Reserves a permanent DOI for the deposit on the main Zenodo service.
+    This does not push any data, but pre-reserves the DOI that will later
+    be used by ``zenodo cache publish``.
+
+    Args:
+        branch (str): Branch to publish.
+    """
+    # Ensure there's a sandbox cache record for this branch
+    if branch is None:
+        branch = get_repo_branch()
+
+    # Get the Zenodo doi, or create one if needed
+    with edit_yaml("zenodo.yml") as config:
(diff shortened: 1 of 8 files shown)


vandalt:
I added an integration test, but it should be rebased on #684 so the test also uses the new passthrough mechanism

src/showyourwork/cli/commands/zenodo.py
```diff
@@ -6,6 +6,41 @@
...
+
+    # Get the Zenodo doi, or create one if needed
+    with edit_yaml("zenodo.yml") as config:
+        zenodo_doi = config["cache"].get(branch, {}).get("zenodo", None)
+    if "SANDBOX_ONLY" in os.environ:
+        # Publish to Sandbox if we're in a test run
+        service = "sandbox"
+    else:
+        service = "zenodo"
+    # TODO: This should be tested
```
HealthyPear: missing a test here? can you add it to this PR?

HealthyPear (changes requested):

src/showyourwork/cli/commands/zenodo.py
vandalt: Done!

HealthyPear:
Could you rebase your PR on the main branch?

vandalt:
Done @HealthyPear !

HealthyPear (approved):

PR #687, opened 2026-03-20
Support including snakemake "rule" files from subdirectories

### LINKED ISSUE #686: [Feature Request]: support snakemake modularization via `include` statements pointing to subdirectories
### Description

Snakemake has a feature to modularize a workflow via `include` statements pointing to other `smk` files, for example in a `srs/rules/` directory (https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#includes).

Currently, this works with showyourwork, but only if the other `smk` files are in the same directory as `Snakefile` and `environment.yml`. Otherwise, there is an error saying that `environment.yml` cannot be found I assume it would work too if an environment was specified directly in the rule. I haven't tried yet. I think the error this happens because of these lines:

https://github.com/showyourwork/showyourwork/blob/7ccdf1ffbdcb0a8193f2f3366ec6ee2bad3bddb1/src/showyourwork/userrules.py#L61-L62

I will look into a fix but I think figuring out the relative path to `environment.yml` should be enough.

_No response_

_No response_

vandalt:
Fixes #686 and adds an integration test for this feature.

### DIFF
--- a/docs/changes/687.feature.rst
+++ b/docs/changes/687.feature.rst
@@ -1,2 +1,2 @@
-showyourwork did support the inclusion of .smk files from subdirectories of the repo via the ``include`` directive.
+showyourwork did not support the inclusion of .smk files from subdirectories of the repo via the ``include`` directive.
 It is now possible to use this feature.
--- a/docs/changes/687.feature.rst
+++ b/docs/changes/687.feature.rst
@@ -1,1 +1,2 @@
-Adds support for Snakemake's ``include`` statement for files in subdirectories.
+showyourwork did support the inclusion of .smk files from subdirectories of the repo via the ``include`` directive.
+It is now possible to use this feature.
--- a/docs/changes/687.feature.rst
+++ b/docs/changes/687.feature.rst
@@ -1,0 +1,1 @@
+Adds support for Snakemake's ``include`` statement for files in subdirectories.
--- a/src/showyourwork/userrules.py
+++ b/src/showyourwork/userrules.py
@@ -1,4 +1,4 @@
-from . import exceptions
+from . import exceptions, paths
 from .config import get_run_type
 from .git import get_repo_branch
 from .logging import get_logger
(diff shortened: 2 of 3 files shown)


src/showyourwork/userrules.py
```diff
@@ -59,6 +59,6 @@ def process_user_rules():
 
         # Ensure we're running in a conda env
         if not ur.conda_env:
-            ur.conda_env = "environment.yml"
+            ur.conda_env = (paths.user().repo / "environment.yml").as_posix()
```
HealthyPear, line 62: I wonder if we should instead throw a warning instead of assuming that this file is enough to run the rule of the user

docs/changes/687.feature.rst
```diff
@@ -0,0 +1 @@
+Adds support for Snakemake's ``include`` statement for files in subdirectories.
```
HealthyPear: I would me more explicit, something like

"*showyourwork* was not supporting the inclusion of Snakefiles in subdirectories from the main Snakefile via the ``include`` directive, now it is possible to use this feature."

HealthyPear (changes requested):

src/showyourwork/userrules.py
vandalt: I think it's fair to assume the default is enough if users don't specify an environment for the rule. That's also how snakemake handles environments.

This is the current behaviour and the PR only makes it consistent when users define the workflow in separate files. From my end-user perspective a warning here would be surprising: I only specified one environment so I do expect all rules to use it unless I specify otherwise.

docs/changes/687.feature.rst
vandalt: Sorry I messed with the commit history while fixing this. Opefully I did not break anything.

docs/changes/687.feature.rst
```diff
@@ -0,0 +1,2 @@
+showyourwork did support the inclusion of .smk files from subdirectories of the repo via the ``include`` directive.
```
HealthyPear: n't :)

HealthyPear (changes requested):

HealthyPear (approved):

PR #688, opened 2026-03-20
Patch snakemake's `onsuccess` hook to include the one from `build.smk` in addition to user's

### LINKED ISSUE #629: [Feature Request]: Support onsuccess and onerror in user's Snakefile
### Description

Currently, showyourwork seems to respect the onstart command from the user's Snakefile, but onsucess and onerror appear to be skipped. I'm assuming this is because they would need to be merged with the equivalent rules somehwere in snakemake (at least for onsuccess, which is in `build.smk`.

The best way to fix this is probably to treat onsucess and onerror as special cases when merging the rules from the user's Snakefil with build.smk. I can look more into this and try to come up with a PR.

My use case for these rules is to do some cleanup after the workflow has run. I guess I could use dummy outputs to create dependencies that always execute the cleanup rules after the last script in my workflow runs, but I think handling `onsuccess` would be cleaner.

Thanks!

vandalt:
By default, snakemake will process all `onsuccess` directives but keep only the last one encountered. In showyourwork, this means the only `onsuccess` kept is the one from `build.smk`, and it will overwrite the one from the user's snakefile. This can be suprising since users don't necessarily know about the internal and can expect the `onsuccess` to work normally, as `onerror` and `onstart` already do.

This PR patches the `onsuccess` hook so that:

- All user rules and files are processed as usual, so that only the last encountered user-defined `onsuccess` is kept. This is consistent with Snakemake
- When we reach `build.smk`, merge the latest user-defined hook with the one from `build.smk`

**Questions/TODO**:

- [x] Currently, `build.smk` only defines an `onsuccess`. Should we also patch `onerror` and `onstart` in case they are used in the future?
- [x] Should we handle not only `build.smk` but all smk files defined within snakemake in case the hook is moved to a different location?
- [x] I still need to write integration tests for this, but it works on a small test repo I did.
- [x] I also need to fix the unit tests

Fixes #629.

### DIFF
--- a/docs/changes/688.feature.rst
+++ b/docs/changes/688.feature.rst
@@ -1,0 +1,2 @@
+Ensure that ``onstart``, ``onsuccess`` and ``onerror`` hook directives created by the user
+get executed normally with Snakemake by merging with showyourwork's internal hooks.
--- a/src/showyourwork/patches.py
+++ b/src/showyourwork/patches.py
@@ -17,7 +17,6 @@ from functools import partial
 from pathlib import Path
 
 from snakemake.caching.local import OutputFileCache as LocalOutputFileCache
-from snakemake.workflow import Workflow
 
 from . import exceptions, paths
 from .logging import ColorizingStreamHandler, get_logger
@@ -102,6 +101,7 @@ def get_snakemake_variable(name, default=None):
 
 
 def patch_snakemake_onsuccess():
+    from snakemake.workflow import Workflow
 
     _onsuccess = Workflow.onsuccess
 
--- a/src/showyourwork/patches.py
+++ b/src/showyourwork/patches.py
@@ -14,6 +14,7 @@ import os
 import time
 import types
 from functools import partial
+from pathlib import Path
 
 from snakemake.caching.local import OutputFileCache as LocalOutputFileCache
 from snakemake.workflow import Workflow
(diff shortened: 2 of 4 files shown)


vandalt:
> Currently, build.smk only defines an onsuccess. Should we also patch onerror and onstart in case they are used in the future?

I went ahead and patched them all. It should not cost a lot (if anything) on compute time and makes the code slightly more future-proof

> Should we handle not only build.smk but all smk files defined within snakemake in case the hook is moved to a different location?

Thinking more about this I think handling all hooks inside showyourwork may lead to unpredictable behaviour so I would tend to stick to the existing issue and handle only `build.smk`. But I don't have a super strong opinion on this.

HealthyPear:
> Thinking more about this I think handling all hooks inside showyourwork may lead to unpredictable behaviour so I would tend to stick to the existing issue and handle only `build.smk`. But I don't have a super strong opinion on this.

The only way one can make sure of this is to write appropriate tests that cover the expected behaviour.
I see that you added a test also for onerror and in general you are also calling onstart but I am not sure how deep we should go

in TestHooksError you do not define a user onerror directive for example

vandalt:
> The only way one can make sure of this is to write appropriate tests that cover the expected behaviour.

You're right. As the tests are currently written any breaking change preventing the user's hooks from running would be caught so we are safe I think.

> in TestHooksError you do not define a user onerror directive for example

It does since `TestHooksError` subclasses `TestHooks` but with `self.hooks = ["onstart", "onerror"]`. This means that `self.customize() -> self.add_hooks()` will call `self.add_hook()` with `"onerror"` instead of `"onsuccess"`.

HealthyPear:
Thanks! (you're in a feature spree lately 😆 )

PR #690, opened 2026-03-23
Support targets in `showyourwork build`

vandalt:
With snakemake, it is possible to specify a target rule or output file to dictate what rule will be executed instead of the default one (at top of the snakefile).

In `showyourwork`, the default rule is `compile`, and since all extra arguments are passed to snakemake it should be possible to override it. Currently this fails due to the preprocess step.

However, the preprocess step does not know about the user rules and should not be affected by these arguments.

This PR removes target rules from the `preprocess` arguments.


HealthyPear:
So, if I understood correctly, with this PR

`showyourwork build my_target`

would be equivalent to 

`snakemake my_target`

skipping the manuscript compilation

?

vandalt:
Yes! I added a changelog entry.

HealthyPear:
Thanks this will be very useful!

Issue #114, opened 2022-06-22
Remove conda dependency?


gusbeane:
Is it possible to remove the conda dependency? I really dislike conda (slow, messes with other python installations). I tried pip installing conda in a virtual environment, but got the following error

```
ERROR: The install method you used for conda--probably either `pip install conda`
or `easy_install conda`--is not compatible with using conda as an application.
If your intention is to install conda as a standalone application, currently
supported install methods include the Anaconda installer and the miniconda
installer.  You can download the miniconda installer from
https://conda.io/miniconda.html.


Conda package manager not found. Is it installed and available in the system PATH?
```

rodluger:
Not really, at least not for the current implementation. ``showyourwork`` is built on top of Snakemake, which makes heavy use of ``conda`` environments to isolate different build steps. It's also the easiest way to get things like ``tectonic`` installed.

I would also love to move past ``conda`` environments and on to something like Docker containers to maximally ensure reproducibility and dependency management. But to be honest I probably won't have the time to implement something like that in the foreseeable future.

gusbeane:
fair enough!

MilesCranmer:
Easiest fix to get improved speed would be to use [mamba](https://mamba.readthedocs.io/en/latest/user_guide/mamba.html) instead of conda, which is a faster rewrite. It uses all the same registries/packages. From their docs:

> mamba is a drop-in replacement and uses the same commands and configuration options as conda. You can swap almost all commands between conda & mamba.

That being said, docker support would be great to have–you could set up the conda environment inside a docker container, there shouldn't be too many changes. Would help keep things separate from user environments!

PR #693, opened 2026-03-24
Add a `--no-conda` CLI flag to run `build` and `tarball` without creating Conda environments

### LINKED ISSUE #692: [Feature Request]: Support running without conda
### Description

In general, I agree it is preferable to run with a dedicated conda environment for reproductibility. However, some environment (e.g. the [main HPC platform in Canada](https://docs.alliancecan.ca/wiki/Anaconda/en)) forbid using conda. For such scenarios, I think it would be nice to be able to run `showyourwork` without having conda available.

I'm testing this right now, but I think it should be enough to add a CLI argument `--no-conda` that removes the `--use-conda` argument from `run_snakemake()`.

Then it's up to the user to have a working environment but someone explicitly passing the `--no-conda` should be aware of that anyway.

_No response_

vandalt:
This `--no-conda` flag simply removes the `--use-conda` flag in `run_snakemake()`. It is then up to the user to run `showyourwork` from a properly configured environment with the required package, and to have a working `tectonic` installation.

I still need to update the docs accordingly and add a changelog entry.

Fixes #692. Also related: #114.

### DIFF
--- a/docs/changes/693.feature.rst
+++ b/docs/changes/693.feature.rst
@@ -1,0 +1,3 @@
+Adds a `--no-conda` flag to the `showyourwork build` and `showyourwork tarball` commands.
+These commands skip the creation of isolated conda environments for the the tex compilation and figure generation.
+Users are expected to have a working Python environment and tectonic installation when using this flag.
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -6,6 +6,14 @@ Installation
 | In order to use |showyourwork|, you'll also need a working installation of the *conda* package manager.
 | We recommend using `miniforge <https://conda-forge.org/download/>`_.
 
+.. note::
+    It is possible to run Snakemake without conda or mamba by
+    passing the ``--no-conda`` flag (see :doc:`Command Line Interface <cli>`).
+    However, we recommend using conda whenever possible to improve reproduciblity.
+    When using showyourwork without conda, you must run it from an environment with
+    all required software dependencies and a working
+    `Tectonic installation <https://tectonic-typesetting.github.io>`_.
+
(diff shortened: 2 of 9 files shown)


vandalt:
I updated the changelog and the docs.

I only mentioned the `--no-conda` in the Installation page (and it gets automatically included in the CLI page). I think this is a good tradeoff: most users will read the Installation page and see this, and it avoids mentioning the caveat "you can actually run without conda" everywhere else in the docs.

For reproducibility and simplicity, I think it is good to encourage users to use conda whenever possible, so this `--no-conda` flag is really just there if needed, for advanced use-cases (e.g. HPC).

Ready for review @HealthyPear :)

HealthyPear (changes requested):
not sure how much of the TODOs in the new test you want to fix in this PR, but for me the minimum is to have a minimum test pass in which we essentially run a simple example equivalent to the quickstart one but without conda to see that it minimally works as we do not (cannot) support any possible user choice in this case

vandalt:
> not sure how much of the TODOs

Oops, those were leftovers. The two cases (with and without conda) are now implemented. In addition to running the workflow, I check if conda environments were created in `.snakemake/conda`.

The "no conda" tests are (finally!) passing on both Ubuntu and Windows. You could mark as safe to test to run the remotes but I don't think it will change much since the github actions still use conda and `--no conda` is only useful for local development.

Thanks!

HealthyPear:
I think this is good to go but I just merged a PR that updated some parts in main which are now conflicting

could you rebase, please?

vandalt:
Done @HealthyPear !

HealthyPear (approved):

PR #694, opened 2026-03-24
Improve Zenodo cache creation and repository setup

vandalt:
Currently, when the repository is setup without the cache, the `showyourwork.yml` config has `cache_on_zenodo` set to `true` anyway, and there is always a warning about "Zenodo cache not yet published for this repository." when running the build.

This PR fixes both issues by:

1. Throwing the warning only when `cache_on_zenodo` is `true`
2. Updating the cookiecutter template and setup function to set `cache_on_zenodo` to `true` only when `--cache` is passed
3. Updating the `showyourwork cache create` command to edit the `showyourwork.yml` file and set `cache_on_zenodo` to `true.

For point 3, I had to add a dependency, `ruamel.yaml`, so that the `showyourwork.yml` could be edited in place while preserving comments and formatting.

I also added unit tests for `edit_yaml` and `edit_yaml_roundtrip` as well as an integration test where the cache is not set as first and is then created.

### DIFF
--- a/docs/changes/694.optimization.rst
+++ b/docs/changes/694.optimization.rst
@@ -1,0 +1,3 @@
+Previously, `showyourwork.yml` always had `cache_on_zenodo: true` by default.
+Now, this is only the case when running `showyourwork setup` with `--cache` or after running `showyourwork cache create`.
+The "Zenodo cache not yet published" warning is also skipped when `cache_on_zenodo` is false.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "graphviz",
     "jinja2",
     "pyyaml",
+    "ruamel.yaml",
     "requests",
     "click",
     "cookiecutter",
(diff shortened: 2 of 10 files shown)


vandalt:
I added a changelog entry. Not sure if this qualifies as bugfix, feature or something else so I went with optimization.

HealthyPear:
weird the remote tests are skipped

vandalt:
Weird indeed. The only info I could find is this:

> Canceling since a higher priority waiting request for showyourwork-remote exists

Maybe two PRs were running the remote tests at the same time? Or something got merge into main which triggered the workflow to run while you marked it safe here?

In any case I'll trigger a local run of the remote tests to check that everything passes.

vandalt:
Some cache tests were failing locally due to #696, but other than that everything seems OK with the remote tests.

HealthyPear:
I guess one of your previous PRs modified test cache too, there is a conflict now hopefully easy to fix!

vandalt:
I just rebased on main. The conflict was only from imports in `test_cache.py` so it should not have changed much compared to before :)

PR #695, opened 2026-03-25
Make Zenodo caching errors fatal

vandalt:
Currently, all Zenodo caching errors are treated as non-fatal and are printed to the logs. While this makes sense for an unstable implementation, I think it is better if we catch specific errors and let the others be raised. This will make problems with the implementation more explicit, and the interface feels stable enough to do this now I think.

This PR only removes the generic `except Exception as e` in the caching patches. All other except clauses are kept. Also, we only attempt to push to the cache if `sandbox.user_is_owner == True`. Otherwise we print a non-fatal warning. Hopefully this is enough to let users run a workflow that they download on GitHub from someone else. I need to test this.

### DIFF
--- a/docs/changes/695.optimization.rst
+++ b/docs/changes/695.optimization.rst
@@ -1,0 +1,3 @@
+All Zenodo caching errors were treated as non-fatal by showyourwork and printed to the logs.
+Unexpected errors will now be raised and prevent the workflow from completing.
+This should make the interface more predictable and easier to debug.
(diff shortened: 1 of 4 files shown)


vandalt:
> Hopefully this is enough to let users run a workflow that they download on GitHub from someone else. I need to test this.

I ran the showyourwork/showyourwork-paper repo and had to handle one issue using an extra try/except (sandbox deposit that no longer exists). Other than that everything seems to work and the integration tests are passing!

I think this is ready for review now.

PR #697, opened 2026-03-28
Fix async function calls in caching patches

### LINKED ISSUE #696: [Bug]: Snakemake v9.18.0 breaks Zenodo caching patches
### Preliminary checks

A recent update in snakemake made `get_outputfiles_and_cachefiles()` async. I think the change is just to add an `await` before the call, but I'll include a version check to avoid breaking older versions of snakemake.

The PR that introduced these changes is https://github.com/snakemake/snakemake/pull/4098, which was merged in release [v9.18.0](https://github.com/snakemake/snakemake/releases/tag/v9.18.0).

### To Reproduce

Run the `test_local()` for `TestCache` with `snakemake>=9.18.0`.

### Logs

_No response_

### Setup information

_No response_

vandalt:
Fixes #696.

- I made `fetch()` async since that was done upstream and added `await`s in it when relevant.
- `store()` was already async so I only had to add an `await`.
- `job_is_cached()` is not async nor are its callers, so I used the `_run_async_safely()` helper for this one.

Instead of handling multiple versions of Snakemake, I felt it was simpler and a reasonable requirement to just enforce `snakemake>=9.18.0`.

### DIFF
--- a/docs/changes/697.bugfix.rst
+++ b/docs/changes/697.bugfix.rst
@@ -1,0 +1,2 @@
+Snakemake updates in version 9.18.1 caused showyourwork caching to raise an error.
+Caching code in showyourwork has been updated to work with the latest version of Snakemake.
--- a/docs/changes/697.bugfix.rst
+++ b/docs/changes/697.bugfix.rst
@@ -1,2 +1,3 @@
 Snakemake updates in version 9.18.1 caused showyourwork caching to raise an error.
 Caching code in showyourwork has been updated to work with the latest version of Snakemake.
+Snakemake version `snakemake>=9.18.1` is now required.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     "click",
     "cookiecutter",
     "packaging",
-    "snakemake>=9.17.3",
+    "snakemake>=9.18.0",
     "pulp",
 ]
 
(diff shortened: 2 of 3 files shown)


HealthyPear:
> Instead of handling multiple versions of Snakemake, I felt it was simpler and a reasonable requirement to just enforce `snakemake>=9.18.0`

that's fine for now (please, update your news fragment with this info!), but I think for the next release we should pin one stable version (then unpin it again for the dev version)

docs/changes/697.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Snakemake updates in version 9.18.1 caused showyourwork caching to raise an error.
+Caching code in showyourwork has been updated to work with the latest version of Snakemake.
```
HealthyPear, line 2: Also add that we are setting a lower limit on snakemake's version

HealthyPear (changes requested):

HealthyPear:
one of 2 PRs was doomed to create a merge conflict about the required snakemake version - could you fix the confict?

HealthyPear:
I believe this PR should fix the error in the CI of #701 and on the remote tests running on `main`

vandalt:
> one of 2 PRs was doomed to create a merge conflict about the required snakemake version - could you fix the confict?

Indeed! The conflict is now fixed :)

HealthyPear (approved):

PR #698, opened 2026-03-28
Enable linting for rule `F821` (undefined names)

vandalt:
The rule was disabled in `ruff.toml` to silence linter errors about `snakemake`. I re-enabled the rules and removed the errors by importing `snakemake` under the `TYPE_CHECKING` condition. The proper way to do this has been updated recently, see https://github.com/snakemake/snakemake/issues/4113 and https://github.com/snakemake/snakemake/pull/4116.

I also updated a missing `rule_name` argument to `_get_draft()`.

I think long term keeping this rule enabled will be beneficial to avoid missing small errors in the codebase.

### DIFF
--- a/docs/changes/698.maintenance.rst
+++ b/docs/changes/698.maintenance.rst
@@ -1,0 +1,2 @@
+Adds a mention of explcit snakemake imports in the showyourwork documentation.
+Enforces ``snakemake>=9.17.3`` to ensure that the internal snakemake imports work.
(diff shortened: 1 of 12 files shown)


vandalt:
I don't think this requires a changelog? Otherwise let me know!

HealthyPear:
> I don't think this requires a changelog? Otherwise let me know!

I think it is not needed

if this change is relevant to a user writing his/her own scripts (not about the specific changes you introduced here, but the why and how you did it) then I would updated the Snakemake page of our docs too to explain what is the new way of importing the snakemake script object (unless I misunderstood your upstream issue)

vandalt:
I updated the docs, required `snakemake>=9.17.3` and added a news fragment accordingly @HealthyPear.

PR #706, opened 2026-04-10
Make dependencies a list, not a string in test_hooks.py

vandalt:
The specs in the docs is clear that each dependencies entry should point to a list. I had used a string in the tests for #688...

This is what caused the error in #670, sorry about that!

I confirm that merging this into 670 allows me to run `test_hooks.py` and get it to pass.


vandalt:
(I don't think this requires a changelog entry since it's only a change in the tests but let me know if you want one @HealthyPear)

PR #716, opened 2026-05-19
Use temporarily micromamba v2.6.0-0 in CI

HealthyPear:
Same temporary workaround suggested in mamba-org/setup-micromamba#306


HealthyPear:
The version pinning of micromammba, which is the main fix originating this PR has uncovered another issue.

I think TestNoConda is corrupting the conda installation and causing a cascade of failures for all subsequent tests

HealthyPear:
They just released a new version of micromamba that is allegedly going to fix the original problem, I suspect that the temp version they released before was breaking something on windows, because on main we did not have any issue

PR #717, opened 2026-05-20
ci: use micromamba 2.6.2-1

### LINKED ISSUE #715: CI setup micromamba fails on Windows due to an upstream bug
See mamba-org/setup-micromamba#306 and issues referenced therein for an explanation.

HealthyPear:
see https://github.com/mamba-org/setup-micromamba/issues/306

Closes #715

### DIFF
--- a/.github/workflows/local_integration_tests.yml
+++ b/.github/workflows/local_integration_tests.yml
@@ -36,6 +36,7 @@ jobs:
       - name: Set up Python and mamba
         uses: mamba-org/setup-micromamba@v3
         with:
+          micromamba-version: '2.6.2-1' # see https://github.com/mamba-org/setup-micromamba/issues/306
           environment-name: test-env
           init-shell: >-
             bash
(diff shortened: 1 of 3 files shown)


HealthyPear:
even with the new version we get the same weird errors on Windows...

thealanjason:
Here are the fixes that are needed:

    Update .github/workflows/local_integration_tests.yml:
        Replace all shell: bash -l {0} with shell: bash
        Wrap all commands with micromamba run -n test-env
        Add --break-system-packages to both pip install commands

    Update .github/workflows/unit_tests.yml:
        Replace all shell: bash -l {0} with shell: bash
        Wrap all commands with micromamba run -n test-env
        Add --break-system-packages to both pip install commands

The complete updated files are shown above. These changes will:

    ✅ Fix the conda environment activation issue (root cause of the original failures)
    ✅ Handle the externally managed environment error
    ✅ Work cross-platform on Ubuntu and Windows

HealthyPear:
Same as for #719, not needed anymore!

PR #719, opened 2026-06-17
Refactor tests workflow to use micromamba correctly

thealanjason:
Updated shell commands to use micromamba for environment management and package installation. Also add --break-system-packages to both pip install commands

See: https://github.com/showyourwork/showyourwork/pull/717#issuecomment-4733227254


HealthyPear:
I think this is not needed anymore: the guys of the setup micromamba action fixed it upstream, indeed e.g. #721 is passing and I just created it from main

thealanjason:
We can safely close this then.

PR #730, opened 2026-07-24
Avoid file-not-found errors due to dangling symlinks

### LINKED ISSUE #729: [Bug]: Renaming txt files stored in src/tex/output via snakemake rule for variable command causes failure due to old name saved in cache
### Preliminary checks

1. I want to save data files in the src/data folder and a txt file, later read by the variable command, in src/tex/output
2. I declared them in a snakemake rule, the corresponding script is required by another plot script
3. The txt file was read by the variable command in tex file
4. Everything worked initially.

I had one snakemake rule, that had three outputs:
I specified them as the following:
```
output:
     var_name_1="src/data/file_name_1",
     var_name_2="src/data/file_name_2",
     var_name_3="src/tex/output/file_name_3"
```

Start of problem:

1. I decided to change the var_name_3 and file name in Snakefile. 
2. I worked when I did `showyourwork build`.
3. For some reason, I tried also `showyourwork clean && showyourwork build`.
4. This caused an error locally because something is still looking for the old filename, which does not exist.
5. I did `ls -la src/tex/output`, which showed `lrwxrwxrwx 1 amajumda amajumda  151 Jul 23 17:46 fit_summary.txt -> /home/amajumda/Dokumente/Software/showyourwork/syw-docs/.showyourwork/cache/f6b60084ad93ba327f5073e82db4a3711dd7e651a4c2c4c3c56d687597e751b5_report.txt`
6. I did `rm src/tex/output/fit_summary.txt`. I think that I also deleted `.showyourowrk` and `.snakemake` locally. This solved the problem locally.
7. However, CI/CD fails.

The error (fit_summary.txt is the old name):

```
/bin/sh: 1: cannot create .showyourwork/tree.txt: Directory nonexistent
  Error: Unable to upload the build logs.
  Error: exec: /bin/sh: 1: cannot create .showyourwork/tree.txt: Directory nonexistent
  
  Error: exec: Traceback (most recent call last):
    File "/home/runner/.conda/bin/showyourwork", line 6, in <module>
      sys.exit(entry_point())
               ~~~~~~~~~~~^^
    File "/home/runner/.conda/lib/python3.14/site-packages/click/core.py", line 1524, in __call__
      return self.main(*args, **kwargs)
             ~~~~~~~~~^^^^^^^^^^^^^^^^^
    File "/home/runner/.conda/lib/python3.14/site-packages/click/core.py", line 1445, in main
      rv = self.invoke(ctx)
    File "/home/runner/.conda/lib/python3.14/site-packages/click/core.py", line 1912, in invoke
      return _process_result(sub_ctx.command.invoke(sub_ctx))
                             ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
    File "/home/runner/.conda/lib/python3.14/site-packages/click/core.py", line 1912, in invoke
      return _process_result(sub_ctx.command.invoke(sub_ctx))
                             ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
    File "/home/runner/.conda/lib/python3.14/site-packages/click/core.py", line 1308, in invoke
      return ctx.invoke(self.callback, **ctx.params)
             ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/.conda/lib/python3.14/site-packages/click/core.py", line 877, in invoke
      return callback(*args, **kwargs)
    File "/home/runner/.conda/lib/python3.14/site-packages/click/decorators.py", line 34, in new_func
      return f(get_current_context(), *args, **kwargs)
    File "/home/runner/.conda/lib/python3.14/site-packages/showyourwork/cli/main.py", line 464, in restore
      commands.cache_restore()
      ~~~~~~~~~~~~~~~~~~~~~~^^
    File "/home/runner/.conda/lib/python3.14/site-packages/showyourwork/cli/commands/cache.py", line 51, in cache_restore
      os.utime(file, (timestamp, timestamp))
      ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  FileNotFoundError: [Errno 2] No such file or directory: 'src/tex/output/fit_summary.txt'
```

### To Reproduce

I also tried solving the problem in remote repo. 
I deleted cache. It worked once but it is not working after that.
I am also generating a dag.

### Logs

_No response_

### Setup information

```python
```

majumdararn:
Fixes #729

Renaming an output file used by a `\variable` command could leave stale symlinks in `src/tex/output`. These dangling symlinks caused subsequent local and GitHub Actions builds to fail with file-not-found errors. This PR makes the build process robust against dangling symlinks.

Solution:
Removed/avoided links when building locally/remotely.


HealthyPear:
Thanks for this fix @majumdararn ! 

Happy to see you found a solution quickly.
Did you happen to understand what is calling the dangling symlink at all? 
Did you also update the filename in the call to `\variable` after changing it in the `Snakefile`? What about in `showyourwork.yml`?
Understanding this would make everything clearer for us and maybe help updating the docs if it is a non-obvious issue.

From the code changes, it seems your fix might be a good guard against this anyway.

To make sure your fix is reproducible, could you add a corresponding test in our test suite?
Since you encountered the bug on both local and most importantly remote pipelines, such test should run also remotely.
Please, take a look at some of the tests already present to get an idea of how to write the smallest test that still is representative of the bug.

I will not yes add the safe-to-test label to run the remote pipeline becuase I discovered that a bug that I though was fixed on the remote CI is not really fixed (see #732), but I will do it asap after the test is available.

majumdararn:
I think,

Locally, the building process is calling the symlink.
Remotely, the cache restore is looking for the symlink.

------------------------------------
I mentioned the file in Snakefile because:

1. I have an analysis pipeline to analyze the data and create a report.
2. I cache the data generated by the analysis pipeline and print a summary of the analysis process, which I call using a variable command.
3. So, I have one script that generates data and creates the txt file to be added. Since this script is not called by a figure, I am defining it using Snakerule.
4. The showyourwork.yml has another figure-generating script, which is dependent on the script in the Snakefile.

Am I doing this correctly?
 
---------------------------------
In general, it is clear to me from the documentation how the cacheing is happening. I cached multiple data but I see only one file (name = rule name). This file does not contain all the data I am cacheing.

---------------------------------------------------
I would be happy to create a test. I need some time.

majumdararn:
Hi @HealthyPear 

I need help.

I created a small test regarding the issue.
The test 
1. creates the repo
2. builds it
3. renames the txt to be called by `\variable`
4. deletes the old file
5. rebuilds

The test happens locally.
Could you please share how could it it done in remote?
Currently, I get the following error with the --remote option.

```
collected 2 items                                                                                                                 
[test-snake-rule] Setting up remote GitHub repo `showyourwork/test-snake-rule`...
Request error 401: Bad credentials

```

HealthyPear:
> Hi @HealthyPear
> 
> I need help.
> 
> I created a small test regarding the issue. The test
> 
> 1. creates the repo
> 2. builds it
> 3. renames the txt to be called by `\variable`
> 4. deletes the old file
> 5. rebuilds
> 
> The test happens locally. Could you please share how could it it done in remote? Currently, I get the following error with the --remote option.
> 
> ```
> collected 2 items                                                                                                                 
> [test-snake-rule] Setting up remote GitHub repo `showyourwork/test-snake-rule`...
> Request error 401: Bad credentials
> ```

dear @majumdararn ,

to be honest I never ran the remote tests from my local machine, always via github
in general this can be done either automatically by calling your branch staging-xxx or by a maintainer via the `safe-to-test` label (which is what we usually do because they are computationally expensive)

I would go for the latter also here if you thhink the test is ready

unfortunately the remote integration test workflow is still failing for other reasons so I need to fix that before

majumdararn:
Hi @HealthyPear,

I don't think that the test is ready, as it probably builds after renaming only locally.
Could you please point me to a test that needs to build remotely after making some change?
I could use it as a reference.

HealthyPear:
OK I managed to fix what I could on the remote integration tests suite, the only error I still see is related to #748 which should not be related to the scope of this PR

For how to write the local and remote version of the test, please take a look at

https://show-your.work/en/latest/integration_tests/#integration-tests

There you should be able to find all required information, but if you do not understand something feel free to ask!

What essentially make a remote test remote is the pytest marker like e.g. in `tests/integration/test_cache.py`

When you think the test is ready let me know and I will label this PR "safe-to-test" so that they can run.

HealthyPear:
Also, please rebase this branch on top of an updated version of main as otherwise some remote existing tests will fail due to e.g. things like #706

Issue #752, opened 2026-09-01
TestZenodo.test_local failing in remote integration suite


HealthyPear:
This popped out recently in the remote integration suite @vandalt , but still not always, maybe it is something caused by unstable network or rate limits

```python
____________________________ TestZenodo.test_local _____________________________

self = <test_zenodo.TestZenodo object at 0x7fd835030710>
caplog = <_pytest.logging.LogCaptureFixture object at 0x7fd834fa9210>

    def test_local(self, caplog):
        """
        Test functionality by creating a new repo, customizing it,
        pushing it to GitHub, and awaiting the article build action to
        complete.
    
        """
        try:
            # Disable screen logging info from showyourwork
            self.disable_logging()
    
            # Always run this first
            self.startup()
    
            # Set up the repo
            self.create_local()
    
            # Customize the repo
            self.customize()
    
            # Commit changes
            self.git_commit()
    
            # Build the article locally
>           self.build_local()

tests/integration/helpers/temp_repo.py:342: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/integration/helpers/temp_repo.py:206: in build_local
    self.run_command(
tests/integration/helpers/temp_repo.py:74: in run_command
    return get_stdout(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../micromamba/envs/test-env/lib/python3.11/site-packages/showyourwork/subproc.py:82: in get_stdout
    return callback(code, stdout, stderr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

code = 1, stdout = ''
stderr = 'Creating conda environment /home/runner/micromamba/envs/test-env/lib/python3.11/site-packages/showyourwork/workflow/e...job execution failed. Look above for error messages\nWorkflowError:\nAt least one job did not complete successfully.\n'

    def callback(code, stdout, stderr):
        if code != 0:
>           raise Exception(stdout + "\n" + stderr)
E           Exception: 
[...]

Issue #751, opened 2026-09-01
[unit tests] test_zenodo_api.py::test_get_id_type - AssertionError


HealthyPear:
Lately this unit-test started failing, but I cannot reproduce it systematically on all machine architectures and all python versions.

@vandalt 

```
_______________________________ test_get_id_type _______________________________

config = {'datasets': {'10.5281/zenodo.6515864': {'destination': 'src/data/TOI640', 'contents': {'README.md': None, 'TOI640b.js...: 'src/data/TOI640/S07.json'}}, 'lightcurves.zip': {'lightcurves': {'README.md': None, 'S06': {...}, 'S07': {...}}}}}}}

    def test_get_id_type(config):
        doi = next(iter(config["datasets"]))
        deposit = Zenodo(doi)
>       assert deposit.get_id_type() == "version"
E       AssertionError: assert 'unknown' == 'version'
E         
E         - version
E         + unknown

tests/unit/test_zenodo_api.py:82: AssertionError
=========================== short test summary info ============================
FAILED tests/unit/test_zenodo_api.py::test_get_id_type - AssertionError: assert 'unknown' == 'version'
  
  - version
  + unknown
```

HealthyPear:
In #753 this did not happen, so it does not seem to be a stable failure, not sure what to think here...

HealthyPear:
I think what happens is that whatever combination of Python/OS runs the test finds a cached "id_type.txt" from a previous run on that machine as per definition of `zenodo.Zenodo.get_id_type`:

```python
    def get_id_type(self):
        """
        Returns the type of a Zenodo `id` ("version" or "concept").

        Caches the result locally.

        """
        cache_file = self.path() / f"{self.deposit_id}" / "id_type.txt"      < --------- it finds it from here

        if cache_file.exists():
            # Restore from cache
            with open(cache_file) as f:
                id_type = f.readline().replace("\n", "")

        else:
            # Try to find a published record (no authentication needed)
            try:
                r = requests.get(f"https://{self.url}/api/records/{self.deposit_id}")
                data = r.json()
            except Exception as e:
                r = None
                data = {"status": "", "message": str(e)}

            if (r is None) or (r.status_code > 204):
                # Either is private or doesn't exist.
                # In any event, don't cache it, as this could change.
                return "unknown"                                                          < --------- not generated here as it is not cached

            elif int(self.deposit_id) == int(data["conceptrecid"]):
                id_type = "concept"
[...]

Issue #748, opened 2026-08-25
Overleaf rate limits during tests


HealthyPear:
Lately some Overleaf tests fail with,

```shell
stderr = 'Fetching Overleaf repo...\n\nfatal: remote error: no git access\nRate-limit exceeded. Please wait a while and try aga..."/home/runner/micromamba/envs/test-env/lib/python3.11/site-packages/showyourwork/overleaf.py", line 103, in callback\n'

Fetching Overleaf repo...

fatal: remote error: no git access
Rate-limit exceeded. Please wait a while and try again.
```

It seems that we hit this issue:

https://docs.overleaf.com/integrations-and-add-ons/git-integration-and-github-synchronization/git-integration#rate-limits

Not sure what we can do here.

What we should probably try to do in the next future is to optimize as much as possible that part of the testing suite (I am mainly talking about the remote integration tests) in order to minimize the use of the Git integratrion.

PR #753, opened 2026-09-01
[remote tests] Fix remaining ValueErrors related to out-of-repo paths 

HealthyPear:
This was caused by a few remote tests not yet updated with the hotfix introduced in #706.

Remaining failing remote tests should fail now due to #752, #748.
I expect also #751


HealthyPear:
OK I can see #748 but not #752, maybe also the latter is a network issue. Anyway the failures related to the scope of this PR seem to have been fixed!

PR #754, opened 2026-09-02
Update unit test `test_zenodo_api.test_get_id_type`

### LINKED ISSUE #751: [unit tests] test_zenodo_api.py::test_get_id_type - AssertionError

HealthyPear:
Clear any stale cache before the test to avoid it to fail on the same machine/runner due to a previous test concurrent run.

Closes #751

### DIFF
--- a/tests/unit/test_zenodo_api.py
+++ b/tests/unit/test_zenodo_api.py
@@ -1,5 +1,6 @@
 import json
 from pathlib import Path
+from unittest.mock import Mock, patch
 
 import pytest
 import snakemake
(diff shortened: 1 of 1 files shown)


HealthyPear:
Nope....there is still something that causes` get_id_type()` to classify the deposit as "unknown".

HealthyPear:
I think the problem is that, even if `config` is defined by a fixture and thus stable ans the test code is, what the test is doing is still "talking" to Zenodo make itself dependent on a live response

this live response could go wrong for a number of reasons (cache state on the specific runner, network issues, etc...) which is why I think this test succeeds randomly on different matrix combinations

perhaps we should just mock the API from zenodo?

HealthyPear:
OK now it works again on all matrix combinations.

What I basically did is the same, but without relying on live Zenodo API calls, just mocking the request via `unittest.mock.Mock` which in my head makes more sense: a unit-test is supposed to test that the function works as expected, it should not depend on potentially unstable live comms with a third-party API.

@vandalt let me know if you think I am breaking something as you are the Zenodo expert!

